PR #14, opened 2023-09-01
Gene info

annacuomo:
Second instance of this pipeline, this script is going to perform processing of expression files and gene info files to generate more SAIGE-QTL input files, specifically:

- pheno_cov_files, <s>one per chromosome + cell type</s> (edit: one per gene + cell type)
- gene_cis_window files, one per gene

[1] combines i) single-cell expression profiles for a given cell type and considering <s>only genes on a certain chromosome</s> each gene separately, ii) single-cell (cell type specific, e.g. expression PCs) + donor covariates (e.g., sex, age), as well as iii) donor information, to match cells to donors of origin.

[2] for each gene, and given a specified window size, create a file specifying the (cis) genomic region we want to test variants in for a specific gene.

@hopedisastro and I are in the process of testing the syntax and input files in a notebook, then will get to running on test but am curious whether at high level, since I am doing a using a lot of loops (over cell types, chromosomes, genes) whether this could be done more efficiently by using batch parallelisation? Maybe @illusional you can have a look or suggest someone else from the software team?

Update: still need input re: batch parallelisation but first modifying the script to:
- use [anndata](https://anndata.readthedocs.io/en/latest/) / [scanpy](https://scanpy.readthedocs.io/en/latest/index.html) as much as possible vs TSV
- only save to TXT at the end (that's the input to SAIGE now) but with only one gene per file

### DIFF
--- a/gene_expression_processing.py
+++ b/gene_expression_processing.py
@@ -26,9 +26,7 @@ import hail as hl
 import pandas as pd
 import scanpy as sc
 
-from cloudpathlib import AnyPath
-
-from cpg_utils import to_path
+from cpg_utils import to_path, Path
 from cpg_utils.hail_batch import (
     dataset_path,
     output_path,
@@ -83,7 +81,7 @@ def get_chrom_celltype_expression(
             f'scrna-seq/CellRegMap_input_files/expression_objects/sce{chromosome}.h5ad'
         )
     ).copy('here.h5ad')
-    expression_h5ad_path = AnyPath(
+    expression_h5ad_path = Path(
         dataset_path(
             os.path.join(expression_files_prefix, cell_type, f'sce{chromosome}.h5ad')
         )
--- a/gene_expression_processing.py
+++ b/gene_expression_processing.py
@@ -58,7 +58,7 @@ def expression_pipeline(
     gene_info_tsv: str,
 ):
     """
-    Run expression prep pipeline
+    Run expression processing pipeline
     """
     gene_info_df = pd.read_csv(gene_info_tsv, sep='\t')
     return gene_info_df
--- a/gene_expression_processing.py
+++ b/gene_expression_processing.py
@@ -22,11 +22,13 @@ import sys
 
 import click
 import hail as hl
+import hailtop.batch as hb
 import pandas as pd
 import scanpy as sc
 
 from cpg_utils import to_path
 from cpg_utils.hail_batch import (
+    copy_common_env,
     dataset_path,
     output_path,
 )
@@ -39,6 +41,8 @@ logging.basicConfig(
     datefmt='%Y-%m-%d %H:%M:%S',
     stream=sys.stderr,
 )
+
+CELLREGMAP_IMAGE = 'australia-southeast1-docker.pkg.dev/cpg-common/images/cellregmap:0.0.1'
 
 
 def filter_lowly_expressed_genes(expression_adata, min_pct=5) -> sc.AnnData:
@@ -182,6 +186,12 @@ def expression_pipeline(
     """
     Run expression processing pipeline
     """
+    sb = hb.ServiceBackend(
+        billing_project=get_config()['hail']['billing_project'],
+        remote_tmpdir=remote_tmpdir(),
+    )
+    batch = hb.Batch('CellRegMap pipeline', backend=sb)
+
     celltype_list = celltypes.split(',')
     chromosome_list = chromosomes.split(',')
     logging.info(f'Cell types to run: {celltype_list}')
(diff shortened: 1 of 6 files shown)


gene_expression_processing.py
```diff
@@ -0,0 +1,242 @@
+#!/usr/bin/env python3
+
+
+"""
+Hail Batch workflow to create gene expression files.
```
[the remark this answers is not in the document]
annacuomo: @illusional what do you think?

gene_expression_processing.py
```diff
@@ -0,0 +1,242 @@
...
+    # get gene chromosome
+    chrom = gene_info_df['chr']
+    # get gene body position (start and end) and add window
+    left_boundary = max(1, int(gene_info_df['start']) - window_size)
+    right_boundary = min(
+        int(gene_info_df['end']) + window_size,
+        hl.get_reference('GRCh38').lengths[chrom],
+    )
+    data = {'chromosome': chrom, 'start': left_boundary, 'end': right_boundary}
+    gene_cis_df = pd.DataFrame(data, index=gene)
```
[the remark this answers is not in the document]
annacuomo: maybe not!

gene_expression_processing.py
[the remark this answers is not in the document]
illusional: Agree with Matt, how long does this script take, if it's a while and you need to run it occasionally, good to turn it into a hail batch workflow.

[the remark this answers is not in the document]
annacuomo: yes especially now we want to write a file per gene i def think hail batch is the way to go. Should I test this (once ready) one say 10 genes and one cell type to verify, or just start converting to Hail Batch? If the latter I can get started, but may need some help

gene_expression_processing.py
```diff
@@ -0,0 +1,272 @@
...
+    - path to dataframe containing gene info, for each gene (=row),
+    specifies chrom, start, end and strand
+
+    Output: expression adata object for only relevant genes
+    """
+
+    # first line is where the file is now,
+    # but (second line) the files will eventually be in the below folder
+    # and split by cell type (at least this all naive B cells only)
+    expression_h5ad_path = to_path(
```
illusional, line 83: You should pass the path into this function as input, it reduces the complexity of this function, and more clearly shows the dependency.

gene_expression_processing.py
```diff
@@ -0,0 +1,272 @@
...
+        dataset_path(
+            f'scrna-seq/CellRegMap_input_files/expression_objects/sce{chromosome}.h5ad'
+        )
+    ).copy('here.h5ad')
```
illusional, line 87: What does `.copy('here.h5ad')` do? It's listed twice on two separate lines

gene_expression_processing.py
```diff
@@ -0,0 +1,272 @@
...
+            for gene in filter_adata.raw.var.index:
+                pheno_cov_job = get_batch().new_python_job(name='creta pheno cov job')
+                copy_common_env(pheno_cov_job)
+                pheno_cov_job.image(CELLREGMAP_IMAGE)
+
+                # pass the output file path to the job, don't expect an object back
+                pheno_cov_job.call(
+                    build_pheno_cov_filename,
+                    gene_name=gene,
+                    cov_df=cov_df,
```
illusional, line 247: Passing in dataframes here is probably a bad idea, it does the serialisation / deserialisation step in pickle, which is likely significantly slower than a regular write + read.

gene_expression_processing.py
```diff
@@ -0,0 +1,272 @@
...
+        _dependent_jobs.append(job)
+
+    for celltype in celltypes.split(','):
+        # get covariates (cell type specific)
+        cov_df = get_celltype_covariates(
+            expression_files_prefix=expression_files_prefix, cell_type=celltype
+        )
+        for chromosome in chromosomes.split(','):
+            # get expression (cell type + chromosome)
+            expr_adata: sc.AnnData = get_chrom_celltype_expression(
```
illusional, line 225: You're completing a lot of loading + compute inside loops.

Can you do all this loading, filtering and deciding what to run outside a loop? Or, is there a way to look at doing this in parallel, because you're paying a sequential and high IO tax lots of times.

gene_expression_processing.py
```diff
@@ -0,0 +1,272 @@
...
+
+import click
+import hail as hl
+import hailtop.batch as hb
+import pandas as pd
+import scanpy as sc
+
+from cpg_utils import to_path
+from cpg_utils.hail_batch import copy_common_env, dataset_path, output_path
+from cpg_workflows.batch import get_batch
```
illusional, line 31: Is there a reason you're using the cpg_workflows get_batch command?
@MattWellie, is this something we should move to cpg-utils?

gene_expression_processing.py
```diff
@@ -0,0 +1,272 @@
...
+    datefmt='%Y-%m-%d %H:%M:%S',
+    stream=sys.stderr,
+)
+
+CELLREGMAP_IMAGE = (
+    'australia-southeast1-docker.pkg.dev/cpg-common/images/cellregmap:0.0.1'
+)
+
+
+def filter_lowly_expressed_genes(expression_adata, min_pct=5) -> sc.AnnData:
```
illusional, line 47: Do other similar scripts specify in whole percentages? I've seen this more commonly as a fraction (float from 0, 1) + it saves you a calculation.

illusional (changes requested):
A couple sporadic comments about the script

gene_expression_processing.py
MattWellie: Yup, happy to look into that

gene_expression_processing.py
MattWellie: One of these lines should be commented out. `copy(XXX)` does the cloudpathlib copy from remote to local (the scanpy library can't read remote paths). If/when this is hail batch'd this will be replaced with a `read_input` call

hopedisastro:
Sorry for the confusion, but can you use this one instead? https://github.com/populationgenomics/saige-tenk10k/pull/24

I worked on the script started in this gene-info branch in a new branch, splitting this script into gene_expression_processing_part1 and part2 scripts. 

The  gene_expression_processing_part2.py script is not in the PR as it has not been finalised. The bare bones of it (memory usage, calling functions) should be roughly what it will be, but the formatting of the input files might change when the new data comes in.

PR #18, opened 2023-09-19
CI fixes

MattWellie:
- adds required libs to requirements
- corrected type hinting (to_path, Path, AnyPath)
- adds return types for AnnData (then tells pylint to stop checking because it's being irrational)


annacuomo (approved):
thank youu

PR #20, opened 2023-10-02
Create gene_expression_processing.py

hopedisastro:


gene_expression_processing.py
```diff
@@ -0,0 +1,263 @@
...
+            # pylint: disable=no-member
+            for gene in filter_adata.var_names:
+                pheno_cov_job = b.new_python_job(name=f'Create pheno cov job for {gene}')
+                pheno_cov_job.storage('35G')
+                pheno_cov_job.cpu(8)
+                pheno_cov_job.image(config['workflow']['driver_image'])
+
+                # pass the output file path to the job, don't expect an object back
+                pheno_cov_job.call(
+                    build_pheno_cov_filename,gene,cov_df,filter_adata,smf_df,output_path(
```
illusional, line 253: Rather than passing the objects directly here `cov_df` `filter_adata`, `smf_df`, pass the paths of the files to the function, and make the function read them. That will dramatically reduce the need for Hail batch to pickle objects, and it'll likely be quicker, because reading the file is quicker than unpickling (usually).

illusional (commented):
I've left a few inline comments, but general rule for PythonJobs, try not to pass objects directly (like a dataframe, or some complex object), pass the `path` to the function, and make the function read it in.

annacuomo:
@hopedisastro, @illusional is it ok to close this PR?

PR #22, opened 2023-10-08
Create LD_prune_plink.py

hopedisastro:


hopedisastro:
This is a subset of the script https://github.com/populationgenomics/saige-tenk10k/blob/main/genotypes_processing.py (lines 290-311). 
The script ran in `main` and failed at this LD prune step (presumably because the Hail image needed to be updated).: https://batch.hail.populationgenomics.org.au/batches/427308/jobs/1

The job costs ~$80 AUD, so instead of re-running all the steps again, I will just run the steps that failed, using the intermediate files outputted already by the job above as input.

PR #25, opened 2023-10-22
Gene expression processing script part2

hopedisastro:
The barebones of the script (memory usage, function calling, batch job initialisation) can be reviewed. But this script is not finalised yet, pending deciding the format of the final input files for the new data.


annacuomo:
@hopedisastro ok to close this PR?

PR #26, opened 2023-12-21
first commit making genotype file for demultiplexing

annacuomo:
Ok ready to go now! The export is resolved and i've removed all logging as it wasn't really necessary here

### DIFF
--- a/extract_bioheart_genos_demux.py
+++ b/extract_bioheart_genos_demux.py
@@ -12,8 +12,6 @@ data so that we can map cells to donors.
 """
 
 # python modules
-import logging
-
 from cpg_utils.hail_batch import (
     dataset_path,
     get_config,
@@ -42,10 +40,8 @@ vds = hl.vds.read_vds(vds_path)
 # split multiallelic loci
 vds = hl.vds.split_multi(vds, filter_changed_loci=True)
 
-# denisfy to matrix table object
+# densify to matrix table object
 mt = hl.vds.to_dense_mt(vds)
-logging.info(f'Number of total loci: {mt.count()[0]}')
-logging.info(f'Number of total samples: {mt.count()[1]}')
 
 # filter out loci & variant QC
 mt = mt.filter_rows(
--- a/extract_bioheart_genos_demux.py
+++ b/extract_bioheart_genos_demux.py
@@ -41,9 +41,7 @@ logging.info(f'Number of total loci: {mt.count()[0]}')
 logging.info(f'Number of total samples: {mt.count()[1]}')
 
 mt = mt.filter_rows(
-    (hl.len(hl.or_else(mt.filters, hl.empty_set(hl.tstr))) == 0)  # QC
-    & (hl.len(mt.alleles) == 2)  # remove hom-ref
-    & (mt.n_unsplit_alleles == 2)  # biallelic (exclude multiallelic)
+    (mt.n_unsplit_alleles == 2)  # biallelic (exclude multiallelic)
     & (hl.is_snp(mt.alleles[0], mt.alleles[1]))  # SNPs (exclude indels)
 )
 
--- a/extract_bioheart_genos_demux.py
+++ b/extract_bioheart_genos_demux.py
@@ -24,7 +24,7 @@ from cpg_utils.hail_batch import (
 import hail as hl
 from hail.methods import export_vcf
 
-DEFAULT_JOINT_CALL_MT = dataset_path(
+BIOHEART_JOINT_CALL_VDS = dataset_path(
     'vds/0c401a0c6aee07549db7d85f88698fbaae4183_990.vds'
 )
 HAIL_IMAGE = get_config()['images']['scanpy']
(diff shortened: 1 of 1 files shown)


extract_bioheart_genos_demux.py
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env python3
+# pylint: disable=import-error,missing-module-docstring,wrong-import-position
+
+__author__ = 'annacuomo'
```
MattWellie, line 4: This should always come after import statements, otherwise pylint (and other linters) will complain

extract_bioheart_genos_demux.py
```diff
@@ -0,0 +1,65 @@
...
+
+vds = hl.vds.read_vds(vds_path)
+
+# split multiallelic loci
+vds = hl.vds.split_multi(vds, filter_changed_loci=True)
+
+# mt = vds.variant_data
+mt = hl.vds.to_dense_mt(vds)
+logging.info(f'Number of total loci: {mt.count()[0]}')
+logging.info(f'Number of total samples: {mt.count()[1]}')
```
MattWellie, line 48: Unless it's absolutely necessary I'd remove these counts. Each call of `mt.count()` forces Hail to carry out the `hl.vds.split_multi(vds, filter_changed_loci=True)` and `hl.vds.to_dense_mt(vds)` operations just to find out how many rows resulted - this is done twice and the results are thrown away. Aim to do `count` operations only after a checkpoint/write, so that all pending operations are resolved - this makes the `count` a computationally free operation

MattWellie (changes requested):
Looks fine, but the computational cost of the debugging/count lines doesn't feel worth it to me

extract_bioheart_genos_demux.py
annacuomo: i was planning on only running this once but fair enough!

MattWellie (approved):
LeggyGregThinksMagnificently

extract_bioheart_genos_demux.py
```diff
@@ -0,0 +1,61 @@
...
+
+# split multiallelic loci
+vds = hl.vds.split_multi(vds, filter_changed_loci=True)
+
+# densify to matrix table object
+mt = hl.vds.to_dense_mt(vds)
+
+# filter out loci & variant QC
+mt = mt.filter_rows(
+    mt.was_split  # biallelic (exclude multiallelic)
```
KatalinaBobowik, line 48: Multiallelic loci should have been split already above (with `hl.vds.split_multi`), so this seems redundant

extract_bioheart_genos_demux.py
```diff
@@ -0,0 +1,61 @@
...
+    & (hl.len(mt.alleles) == 2)  # remove hom-ref
+    & (hl.is_snp(mt.alleles[0], mt.alleles[1]))  # SNPs (exclude indels)
+)
+mt = hl.variant_qc(mt)
+
+# common variants only
+cv_mt = mt.filter_rows(mt.variant_qc.AF[0] > 0.05)
```
KatalinaBobowik, line 55: This should be for the variant ([1]), so `mt.variant_qc.AF[1] > 0.05`

extract_bioheart_genos_demux.py
```diff
@@ -0,0 +1,61 @@
...
+
+# remove fields not in the VCF
+cv_mt = cv_mt.drop('gvcf_info')
+
+# export to plink common variants only for demultiplexing
+export_vcf(cv_mt, cv_demux_vcf_path)
```
KatalinaBobowik, line 61: I haven't exported many VCFs but is this not `hl.export_vcf`?

extract_bioheart_genos_demux.py
annacuomo: That's what I thought as well! But when I tested in my notebook it was the was_split line that removed most of the loci, so I assume the line below was not dealing well with these? See screenshot:
Screen Shot 2023-12-29 at 4 43 19 pm

am i missing something?

extract_bioheart_genos_demux.py
```diff
@@ -0,0 +1,61 @@
...
+BIOHEART_JOINT_CALL_VDS = dataset_path(
+    'vds/0c401a0c6aee07549db7d85f88698fbaae4183_990.vds'
+)
+HAIL_IMAGE = get_config()['images']['scanpy']
+
+# read hail matrix table object (WGS data)
+init_batch()
+
+vds_path = BIOHEART_JOINT_CALL_VDS
+cv_demux_vcf_path = output_path('demux_vcf_common_variants.vcf.bgz')
```
KatalinaBobowik, line 36: nit but I would put this above L61, when you're exporting

extract_bioheart_genos_demux.py
annacuomo: I hadn't used this before but it's the same syntax as export_plink (which I have used in the past):

it's imported as `from hail.methods import export_plink, export_vcf` and then can be run as I did

Ref: https://hail.is/docs/0.2/methods/impex.html#id2

extract_bioheart_genos_demux.py
annacuomo: ok!

extract_bioheart_genos_demux.py
KatalinaBobowik: I think it's just saying that that locus was indeed split, even though now it's biallelic. So before the split it was multiallelic, but a split happened to it when `hl.vds.split_multi` was run

extract_bioheart_genos_demux.py
KatalinaBobowik: oh neat! I hadn't seen that one

extract_bioheart_genos_demux.py
annacuomo: ah oops I always get this wrong! I think i am getting the use of filter_rows wrong as well, i actually want to exclude / filter out the rare ones, so I want `mt.filter_rows(mt.variant_qc.AF[1] < 0.05)` (smaller than, vs larger than) I think!!

KatalinaBobowik: `mt.filter_rows(mt.variant_qc.AF[1] > 0.05)` will _keep_ all variants that are above a frequency of 0.05, so rare variants below that threshold will be removed.

annacuomo: yes sorry i confused myself again, default is keep=True so your solution is correct, I'll make that change

extract_bioheart_genos_demux.py
annacuomo: yes so now it is keeping all of these formerly multiallelic loci right, because it sees them as biallelic? Still I need to is_split filter to be the reverse, do you know what the syntax is for that?

annacuomo:
more fixes in new PR:  https://github.com/populationgenomics/saige-tenk10k/pull/28

PR #28, opened 2023-12-29
Demux2

annacuomo:
ok should've made those new changes in this new PR, I am just uncertain of the syntax at L47, where I want to exclude previously multiallelic loci!

### DIFF
--- a/extract_bioheart_genos_demux.py
+++ b/extract_bioheart_genos_demux.py
@@ -4,8 +4,13 @@
 """
 This script will
 
-- extract common variants
+- extract common, biallelic SNPs
 - export as VCF
+
+this excludes non-variable loci since they will not help
+in differentiating individuals, multi-allelic SNPs, as they
+are more likely to be artefact and harder to find in RNA data,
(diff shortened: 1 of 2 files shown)


extract_bioheart_genos_demux.py
```diff
@@ -45,17 +44,18 @@
 
 # filter out loci & variant QC
 mt = mt.filter_rows(
-    mt.was_split  # biallelic (exclude multiallelic)
+    ~(mt.was_split)  # biallelic (exclude multiallelic)
```
KatalinaBobowik, line 47: that's cool! I was thinking `mt.filter_rows(mt.was_split == False)`, but that looks cleaner

KatalinaBobowik (approved):
LGTM!

extract_bioheart_genos_demux.py
KatalinaBobowik: could you maybe just add in a comment about not wanting the previous multiallelics due to sequencing artefacts?

MattWellie (approved):

PR #30, opened 2024-01-17
Handle real files in batch

MattWellie:
Not a million percent sure here, but this should pass files between jobs, and if the output paths specified on the command line start with `gs://` it will also attempt to store them in GCP

### DIFF
--- a/saige_assoc.py
+++ b/saige_assoc.py
@@ -237,16 +237,26 @@ def association_pipeline(
     fit_null_job = batch.new_job(name='fit null')
     fit_null_job.image(SAIGE_QTL_IMAGE)
     fit_null_job.storage(fit_null_mem)
+    fit_null_job.declare_resource_group(
+        output={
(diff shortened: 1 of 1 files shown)


annacuomo:
> Not a million percent sure here, but this should pass files between jobs, and if the output paths specified on the command line start with `gs://` it will also attempt to store them in GCP

awesome, can you clarify why you only need `declare_resource_group` for the first job and not the following ones? just curious :)

MattWellie:
> can you clarify why you only need `declare_resource_group` for the first job and not the following ones

I was kinda basing this on the code that's here - the first 'output file path' is actually just the root folder, the code will generate a `OUTPUT.rda` file based on the `OUTPUT` you provided. Hail needs a resource group to pull them back out, where you tell it the various extensions that will be generated after OUTPUT, and then it can collect all the individual files.

That might also be true for the next two outputs, but I don't know which extensions are generated

PR #29, opened 2024-01-11
Association pipeline

annacuomo:
Edit: this ran successfully in test: https://batch.hail.populationgenomics.org.au/batches/432054

Reminder that at this stage this was simply testing the installation / basic usage of SAIGE-QTL on GCP, as well as how to access / write to the buckets (so probably not super meaningful from a scientific perspective yet).

Can we approve so I can build on this to run on real data? 

---

SAIGE-QTL is now ready to run, so would like to start setting this part of the pipeline up for testing maybe with test data (there is some as part of the method in the github repo, see below).

What is the easiest way to install the newest version of https://github.com/weizhou0/qtl? 
Can we just update the Docker image here? https://github.com/populationgenomics/images/tree/main/images/saige-qtl

And how can the example data from ext data (in input and output here: https://github.com/weizhou0/qtl/tree/main/extdata) be added to the image? I am currently setting them as defaults for testing. I am following the docs from https://weizhou0.github.io/SAIGE-QTL-doc/ atm.

Any help on how to get started + syntax of actually running this (are my current batch jobs sensible?) would be awesome @MattWellie @vivbak 

@KatalinaBobowik @hopedisastro FYI, SAIGE-QTL is the method we'll use for association, so any comments on the science, how to add more comments to make clearer etc would also be appreciated! Note that each step is only run for one gene, so when doing this for real we'll need some way of looping / parallelising across cell types + genes and then also a final part summarising all results!

PR #32, opened 2024-01-17
Association file only

annacuomo:
New version of https://github.com/populationgenomics/saige-tenk10k/pull/29 without all the mismatches with main / changes to other scripts due to how old that branch was.

### DIFF
--- a/saige_assoc.py
+++ b/saige_assoc.py
@@ -14,6 +14,7 @@ Hail Batch workflow to perform association tests using SAIGE-QTL
 To run:
 
 analysis-runner \
+    --config saige_assc.toml \
     --description "SAIGE-QTL association pipeline" \
     --dataset "bioheart" \
     --access-level "test" \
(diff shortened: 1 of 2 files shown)


saige_assoc.py
```diff
@@ -0,0 +1,312 @@
...
+        variance_ratio_path=fit_null_job.output['varianceRatio'],
+    )
+    run_sv_assoc_job.command(cmd)
+    if sv_output_path.startswith('gs://'):
+        batch.write_output(run_sv_assoc_job.output, sv_output_path)
+
+    # step3 (gene-level pvalues)
+    get_gene_pvals_job = batch.new_job(name='gene level pvalues')
+    get_gene_pvals_job.image(SAIGE_QTL_IMAGE)
+    get_gene_pvals_job.storage(get_gene_pvals_mem)
```
MattWellie, line 296: A fussy little nit here, please be careful semantically about `mem` and `storage` - one is RAM, one is hard-drive capacity. The hail batch job commands `.memory()` and `.storage()` address very different needs, so this variable has a misleading name

MattWellie (approved):

PR #34, opened 2024-01-18
start adding to this for real analysis

annacuomo:
Update this ran successfully all the way in `test` (https://batch.hail.populationgenomics.org.au/batches/434527), but using genotype and sample cov files copied from main (https://github.com/populationgenomics/transfer-private/blob/main/bioheart/copy_inputs_main_to_test.sh) and expression AnnData and cell cov files also a subset of main objects (https://github.com/populationgenomics/transfer-private/blob/main/bioheart/transfer_anndata_saige_pipeline.sh and https://github.com/populationgenomics/transfer-private/blob/main/bioheart/transfer_celltype_covs_saige_pipeline.sh) so should be ready to merge from my perspective.

-----

SAIGE-QTL runs successfully on toy example files, but am pausing it temporarily while I work out how to build actual input files, so that I can test.

Current status:

- step 1 is finished and merged, and can be run in main (https://github.com/populationgenomics/saige-tenk10k/blob/main/get_genotype_vcf.py)
- step 2 is essentially finished, just being tested from: https://github.com/populationgenomics/saige-tenk10k/blob/expression-process/get_anndata.py 
- sample covariates (sex + age), which will feed into step 2 is also finished and merged: https://github.com/populationgenomics/saige-tenk10k/blob/main/get_sample_covariates.py

Once step 2 is finalised, hopefully this will be just a matter of extending: https://github.com/populationgenomics/saige-tenk10k/blob/main/saige_assoc.py to include looping over genes and celltypes to run systematically for everything


saige_assoc.py
```diff
@@ -200,123 +238,255 @@ def apply_job_settings(job, job_name: str):
...
+
+    Returns:
+        Tuple: (Job | None, ResourceGroup)
+
+    """
+    if to_path(null_output_path).exists():
+        return None, get_batch().read_input_group(
+            **{
+                'rda': f'{null_output_path}.rda',
+                'varianceRatio': f'{null_output_path}.varianceRatio.txt',
```
MattWellie, line 273: is this definitely the right extension?

saige_assoc.py
```diff
@@ -200,123 +238,255 @@ def apply_job_settings(job, job_name: str):
...
+        )
+    gene_job = get_batch().new_job(name="saige-qtl")
+    gene_job.image(image_path('saige-qtl'))
+    apply_job_settings(gene_job, 'fit_null')
+
+    # create output group for first command in gene job
+    gene_job.declare_resource_group(
+        output={
+            'rda': '{root}.rda',
+            'varianceRatio': '{root}.varianceRatio.txt',
```
MattWellie, line 284: same here, if the files aren't created with `.txt` extensions we should probably drop them here

MattWellie (approved):

saige_assoc.py
annacuomo: I think so! 

See for example the example output in the SAIGE-QTL repo: https://github.com/weizhou0/qtl/blob/main/extdata/output/example_binary_1.1.4.varianceRatio.txt

And also the documentation:
>  less -S ./output/nindep_100_ncell_100_lambda_2_tauIntraSample_0.5_gene_1.varianceRatio.txt
https://weizhou0.github.io/SAIGE-QTL-doc/docs/step1.html

What makes you think it is not?

saige_assoc.py
annacuomo: see above, I think they should be!

MattWellie, line 284:
```suggestion
            'varianceRatio.txt': '{root}.varianceRatio.txt',
```

saige_assoc.py
MattWellie, line 273:
```suggestion
                'varianceRatio.txt': f'{null_output_path}.varianceRatio.txt',
```

MattWellie (changes requested):

saige_assoc.py
```diff
@@ -91,28 +102,31 @@ def build_fit_null_command(
...
     min_maf: float = 0,
     min_mac: int = 5,
     loco_bool: str = 'FALSE',
     n_markers: int = 10000,
     spa_cutoff: int = 10000,
 ):
-    """Build SAIGE command for running single variant test
+    """
+    n.b. I have not edited this method docstring
+            ... but I have definitely edited the method
```
MattWellie, line 127: 😆

saige_assoc.py
```diff
@@ -200,123 +238,255 @@ def apply_job_settings(job, job_name: str):
...
+                    manage_concurrency_for_job(null_job)
+
+                # step 2 (cis eQTL single variant test)
+                step2_job, step2_output = build_run_single_variant_test_command(
+                    output_path=output_path(f'output_files/{celltype}_{gene}_cis'),
+                    vcf_file=vcf_file_path,
+                    chrom=chromosome,
+                    cis_window_file=cis_window_path,
+                    gmmat_model_path=null_output['rda'],
+                    variance_ratio_path=null_output['varianceRatio'],
```
MattWellie, line 453:
```suggestion
                    variance_ratio_path=null_output['varianceRatio.txt'],
```

MattWellie (approved):

MattWellie:
Just trying to piece two things together to make sense of this:

https://hail.is/docs/batch/api/batch/hailtop.batch.batch.Batch.html#hailtop.batch.batch.Batch.write_output
```
write_output()
Parameters:
- resource (Resource) – Resource to be written to a file.
- dest ([str] – Destination file path. For a single ResourceFile, this will simply be dest. For a ResourceGroup, dest is the file root and each resource file will be written to {root}.identifier where identifier is the identifier of the file in the ResourceGroup map.
```

And ResourceGroup: https://hail.is/docs/batch/api/resource/hailtop.batch.resource.ResourceGroup.html#hailtop.batch.resource.ResourceGroup
```Class representing a mapping of identifiers to a resource file.```

As I understand it, if a resource group is declared:
```
job.declare_resource_group(
        output={
            'ext1': '{root}.1',
            'ext2': '{root}.2',
        }
    )
```

then these can be used as `job.output['ext1']` to point to the temp file in the batch `BATCH_TMP/output.1`

When these files are written out into permanent storage, 
```
batch.write_output(job.output, 'path')
```

will write the files `path.ext1` and `path.ext2`, taking the left-hand side of the resource group as the extension. I think that's why the files were being written out with just the extension `varianceRatio`, without the additional `.txt`

saige_assoc.py
annacuomo: ok updated this as well, although I'll have a new PR to deal with all of these parameters in a cleaner way coming soon

PR #36, opened 2024-01-31
begin new file making vcf files for assoc

annacuomo:
Ok so this script will substitute genotype_processing, as I now will use the VDS object instead of the MT, and I don't need to do all of the cleaning of samples that the old file required.

I have started off the same way as the `extract_bioheart_genos_demux.py` script, but have added 1) that i want one vcf file per chromosome, and 2) the indexing with bcftools

i am sure i made many mistakes and maybe the first export should also be a job? Comments welcome :)

### DIFF
--- a/get_genotype_vcf.py
+++ b/get_genotype_vcf.py
@@ -39,13 +39,13 @@ def main(chromosomes):
     for chromosome in chromosomes:
 
         # consider only relevant chromosome
-        vds = hl.vds.filter_chromosomes(vds, keep=chromosome)
+        chrom_vds = hl.vds.filter_chromosomes(vds, keep=chromosome)
 
         # split multiallelic loci
-        vds = hl.vds.split_multi(vds, filter_changed_loci=True)
+        chrom_vds = hl.vds.split_multi(chrom_vds, filter_changed_loci=True)
 
         # densify to matrix table object
-        mt = hl.vds.to_dense_mt(vds)
+        mt = hl.vds.to_dense_mt(chrom_vds)
 
         # filter out loci & variant QC
         mt = mt.filter_rows(
--- a/get_genotype_vcf.py
+++ b/get_genotype_vcf.py
@@ -1,0 +1,66 @@
+#!/usr/bin/env python3
+# pylint: disable=import-error
+
+"""
+This script will
+
+- extract common variants
+- export as VCF (one per chrom)
+
+this will be used as input for the
+SAIGE-QTL association pipeline.
+"""
+
+# python modules
+from cpg_utils.hail_batch import (
+    dataset_path,
+    get_config,
+    init_batch,
+    output_path,
+)
+
+import hail as hl
+from hail.methods import export_vcf
+
+
+BIOHEART_TOB_VDS = dataset_path('.vds')
+HAIL_IMAGE = get_config()['images']['scanpy']
+CHROMOSOMES = 'chr22'
+
+# read hail matrix table object (WGS data)
+init_batch()
+
+vds_path = BIOHEART_TOB_VDS
+
+vds = hl.vds.read_vds(vds_path)
+
+for chromosome in CHROMOSOMES:
+
+    # consider only relevant chromosome
+    vds = hl.vds.filter_chromosomes(vds, keep=chromosome)
+
+    # split multiallelic loci
+    vds = hl.vds.split_multi(vds, filter_changed_loci=True)
+
+    # densify to matrix table object
+    mt = hl.vds.to_dense_mt(vds)
+
+    # filter out loci & variant QC
+    mt = mt.filter_rows(
+        ~(mt.was_split)  # biallelic (exclude multiallelic)
+        & (hl.len(mt.alleles) == 2)  # remove hom-ref
+        & (hl.is_snp(mt.alleles[0], mt.alleles[1]))  # SNPs (exclude indels)
+    )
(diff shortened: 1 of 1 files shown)


get_genotype_vcf.py
```diff
@@ -0,0 +1,85 @@
...
+"""
+
+# python modules
+from cpg_utils.hail_batch import (
+    dataset_path,
+    get_batch,
+    get_config,
+    init_batch,
+    output_path,
+)
```
MattWellie, line 21:
```suggestion
from cpg_utils.config import get_config
from cpg_utils.hail_batch import dataset_path, get_batch, init_batch, output_path
```

get_genotype_vcf.py
```diff
@@ -0,0 +1,85 @@
...
+@click.option('--chromosomes', help=' eg chr22')
+@click.command()
+def main(chromosomes):
+    """
+    Run associaTR processing pipeline
+    """
+
+    init_batch()
+
+    vds_path = BIOHEART_TOB_VDS
```
MattWellie, line 41: do you want this to be configurable at any point? NW if this is a one-off script

get_genotype_vcf.py
```diff
@@ -0,0 +1,85 @@
...
+    vds = hl.vds.read_vds(vds_path)
+
+    for chromosome in chromosomes:
+
+        # consider only relevant chromosome
+        vds = hl.vds.filter_chromosomes(vds, keep=chromosome)
```
MattWellie, line 47:
```suggestion
        chrom_vds = hl.vds.filter_chromosomes(vds, keep=chromosome)
```
this will overwrite the vds object, so you need to rename it (otherwise the next loop will find no data remaining on its relevant chromosome)

get_genotype_vcf.py
```diff
@@ -0,0 +1,85 @@
...
+        export_vcf(cv_mt, cv_vcf_path)
+
+        # add index file (.csi) using bcftools
+        vcf_input = get_batch().read_input(cv_vcf_path)
+        bcftools_job = get_batch().new_job(name='index vcf')
+        bcftools_job.image(BCFTOOLS_IMAGE)
+        bcftools_job.cpu(4)
+        bcftools_job.storage('15G')
+        bcftools_job.command(f"bcftools index -c {vcf_input} -o {cv_vcf_path}.csi")
+
```
MattWellie, line 80:
```suggestion
        bcftools_job.command(f"bcftools index -c {vcf_input} -o {bcftools_job.csi}")
        get_batch().write_output(bcftools_job.csi, f'{cv_vcf_path}.csi')
```

afaik bcftools won't be able to write directly to the GCP path, so you need to create a temp file then copy to GCP separately

get_genotype_vcf.py
MattWellie: this new chrom-specific name will need to be used in the next few lines as well

MattWellie (changes requested):
Almost nailed it!

get_genotype_vcf.py
annacuomo: mm i did think that in some ways this is similar to the demux script and maybe i could make it more general? I might end up running this many times, but always on the BioHEART TOB joint call object i believe

get_genotype_vcf.py
annacuomo: the lack of available emojis is letting me down -- yes of course i'm dumb, will do this.

Should this be a method / function i call for every chromosome instead though?

get_genotype_vcf.py
annacuomo: ah yes i see! Is that the syntax to get the job's output?? Just job name + .csi?

get_genotype_vcf.py
MattWellie: On the basis that the joint-calls will change name each time, I think providing a name as a command line arg would be good. You don't want a PR to change each input

get_genotype_vcf.py
MattWellie: It depends how complicated what you're doing for each chromosome is.. This is quite short and linear, so you are probably fine as it is

get_genotype_vcf.py
MattWellie: every time you use `job.something` Hail reserves a name in the batch temp space for `something`, which can be accessed after to pass into other jobs or to copy into permanent storage. In the hail batch tutorials I'm pretty sure they use `job.output` each time, but any name will work.

MattWellie (approved):

get_genotype_vcf.py
```diff
@@ -0,0 +1,91 @@
...
+import click
+import hail as hl
+from hail.methods import export_vcf
+
+
+BCFTOOLS_IMAGE = get_config()['images']['bcftools']
+
+# inputs:
+@click.option('--vds-name', help='e.g., vds/1-0.vds')
+@click.option('--chromosomes', help='e.g., chr22')
```
MattWellie, line 38:
```suggestion
@click.argument('--chromosomes', help='e.g., chr22 chrX ', nargs=-1)
```

alternative is to switch to click arguments. This would make your command line call something like

```
python3 get_genotype_vcf.py --vds-name vds/1-0.vds chr1 chr2 chr22
```

(no need to use `--chromosomes` at all)

annacuomo: what does `nargs=-1 `do?

PR #38, opened 2024-01-31
Add to make genotypes

annacuomo:
I decided this script was small enough that I could include making of another needed input file for the pipeline (a plink file containing only a subset of common variants across all chromosomes, used for some internal cheaper estimations)

first, i have suggested changes to provide the vds version instead of name as an input, so that my outputs can also be versioned in the same way.

second, I have added a check for the vcf file existence -- the entire thing only happens if the file does not already exist

third, I have now added the VRE plink file making but i think it's very ugly and repetitive


MattWellie (approved):
Spicy 🌶️

PR #46, opened 2024-02-15
Update saige_assoc.py

MattWellie:
Quick change to how the strings are generated and manipulated from Path/Glob search


annacuomo (commented):
great thank you!

PR #48, opened 2024-02-16
maybe this?

MattWellie:
The [export_plink docs](https://hail.is/docs/0.2/methods/impex.html#hail.methods.export_plink) say that it accepts a `Locus`, or any `Struct` that implements a contig + position. This change (untested) takes the MT, creates a new version of `locus` without the `chr` prefix, uses that as a key, then export_plink is called. 

Not sure if a similar trick is possible with a VCF export - I doubt it, that would mean a VCF is written where the 'contigs' in the variant rows don't match anything in the headers


annacuomo:
@MattWellie is it ok to close this PR?

PR #49, opened 2024-02-16
remove chr from geno files

annacuomo:
create methods to remove the chr in the chrom column for both the vcf (adding to the bcftools job) and the bim files


MattWellie (approved):
Approval powered by Matt

PR #50, opened 2024-02-18
overwrite vcf

annacuomo:
The operation of removing "chr" from genotype files had not worked.

For the VCF at least, I think it's a matter of writing the new file (after removing chr) to an output properly

### DIFF
--- a/get_genotype_vcf.py
+++ b/get_genotype_vcf.py
@@ -211,14 +211,14 @@ def main(
         # # check existence of bim file separately
         # if not can_reuse(f'{vre_plink_path}.bim'):
         # remove chr using awk
-        plink_input = get_batch().read_input(vre_plink_path)
+        plink_input_bim = get_batch().read_input(f'{vre_plink_path}.bim')
         remove_chr_job = get_batch().new_python_job(name='remove chr from plink bim')
         remove_chr_job.cpu(4)
         remove_chr_job.storage('15G')
         # remove chr
-        remove_chr_job.call(remove_chr_from_bim, plink_input.bim, plink_input.bim)
+        remove_chr_job.call(remove_chr_from_bim, plink_input_bim, plink_input_bim)
         logging.info('chr removed from bim')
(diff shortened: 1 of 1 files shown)


get_genotype_vcf.py
```diff
@@ -74,8 +74,18 @@ def remove_chr_from_bim(input_bim, output_bim):
 @click.option('--cv-maf-threshold', default=0.01)
 @click.option('--vre-mac-threshold', default=20)
 @click.option('--vre-n-markers', default=2000)
+@click.option('--exclude-multiallelic', default=False)
+@click.option('--exclude-indels', default=False)
```
MattWellie, line 78: A better click argument here might be `is_flag=True`

MattWellie (approved):

get_genotype_vcf.py
```diff
@@ -127,12 +159,15 @@ def main(vds_version, chromosomes, cv_maf_threshold, vre_mac_threshold, vre_n_ma
             bcftools_job.image(BCFTOOLS_IMAGE)
             bcftools_job.cpu(4)
             bcftools_job.storage('15G')
-            # remove chr
+            # now remove "chr" from chromosome names
+            rename_file = create_rename_chr(rename_file)
             bcftools_job.command(
-                f'bcftools annotate --remove-chrs {vcf_input} | bgzip > {vcf_input}'
+                f'bcftools annotate --rename-chrs {rename_file} {vcf_input} -o {bcftools_job.vcf}'
```
MattWellie, line 165:
```suggestion
                f'for num in  {1..22} X Y; do echo "chr${num} ${num}" >> chr_update.txt; done && bcftools annotate --rename-chrs chr_update.txt {vcf_input} -o {bcftools_job.vcf}'
```

MattWellie: Not sure this is the best one-liner, but the command 
```for num in  {1..22} X Y; do echo "chr${num} ${num}" >> chr_update.txt; done```
will generate the file you want as a local file

PR #58, opened 2024-02-27
Keep bcftools output compressed

MattWellie:
This needs to be one block command - writing the un-compressed output to a temp is blowing up storage requirements


annacuomo (approved):
Good catch! Can you black or should I

PR #61, opened 2024-02-28
Remove doubled suffix

MattWellie:
similar https://github.com/populationgenomics/production-pipelines/blob/main/cpg_workflows/stages/gatk_sv/gatk_sv_multisample_2.py#L606-L622


annacuomo (approved):
great thank you!!

PR #62, opened 2024-02-28
add cpu and storage to batch jobs

annacuomo:
Specify cpus and storage for jobs? Set to random values for now, or could add as args?


MattWellie (approved):
LGTMatt

PR #63, opened 2024-02-28
Make GCP temp object and pass location to python Jobs

MattWellie:
Hail is (or appears to be) having scaling issues when passing large data objects into batch jobs. This change avoids that behaviour by instead making a copy in GCP of the relevant AnnData object at the time the jobs are being created, then passes the GCP path to the temp file. 

Each job independently copies and opens the file, moving the burden from hail batch to the individual pythonJobs


annacuomo (approved):
ill give it a try!

PR #65, opened 2024-02-28
Use Hail's hadoop_open method to write

MattWellie:


annacuomo (approved):
reproducing how this works fine in `test` when testing only a handful of genes:
https://batch.hail.populationgenomics.org.au/batches/434132/jobs/1

and confirming this latest change does not affect that

PR #67, opened 2024-02-29
Export rare variants vcf

annacuomo:
Update: removed most of the repetition by creating an `add_remove_chr_and_index_job` function as suggested by @jmarshall below. Code ran successfully in test (https://batch.hail.populationgenomics.org.au/batches/434456/jobs/1) and is ready for merging from my perspective!

-------------- 

There's quite a bit of repetition in this code, when adding the rare variant VCF making the procedure is essentially the same?

- export to VCF
- index VCF
- remove the "chr" notation from the chromosomes
- save

Any suggestions to make this cleaner or is it ok?

### DIFF
--- a/get_genotype_vcf.py
+++ b/get_genotype_vcf.py
@@ -113,6 +113,7 @@ def can_reuse(path: str):
         return True
     return False
 
+
 def add_remove_chr_and_index_job(vcf_path):
     """
     Reads a VCF file, it creates an .csi index file and
@@ -150,9 +151,7 @@ def add_remove_chr_and_index_job(vcf_path):
     logging.info('CV VCF rename/index jobs scheduled!')
 
     # save both output files
-    get_batch().write_output(
-        bcftools_job.output, vcf_path.removesuffix('.vcf.bgz')
-    )
+    get_batch().write_output(bcftools_job.output, vcf_path.removesuffix('.vcf.bgz'))
 
 
 def remove_chr_from_bim(input_bim: str, output_bim: str, bim_renamed: str):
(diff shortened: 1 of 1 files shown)


jmarshall:
It looks basically fine to me. If you did want to get rid of that repetition (which isn't the end of the world but would be a bit better using a function), I'd suggest turning the stuff inside the `if not BLAH_index_file_existence_outcome:` if statements into a function with parameters something like

```python
def add_remove_chr_and_index_job(vcf_path)
```

which would be either `cv_vcf_path` or `rv_vcf_path` as appropriate — hmmm I was expecting it to need more parameters, but that's quite clean!

annacuomo:
> It looks basically fine to me. If you did want to get rid of that repetition (which isn't the end of the world but would be a bit better using a function), I'd suggest turning the stuff inside the `if not BLAH_index_file_existence_outcome:` if statements into a function with parameters something like
> 
> ```python
> def add_remove_chr_and_index_job(vcf_path)
> ```
> 
> which would be either `cv_vcf_path` or `rv_vcf_path` as appropriate — hmmm I was expecting it to need more parameters, but that's quite clean!

Ok I just attempted to apply your suggestion, testing here: https://batch.hail.populationgenomics.org.au/batches/434447/jobs/1

I never fully understand when to use str vs paths, and when to set jobs running either so this may be complete garbage

jmarshall:
I was going to say “looks good but I've only eye-balled it, not actually run it”. However I see from your last comment that you have now run this PR's branch — excellent :+1:

jmarshall (approved):

PR #68, opened 2024-02-29
Rare variants group file making

annacuomo:
Ready to review now.

Ok finally successfully ran in test: https://batch.hail.populationgenomics.org.au/batches/478829/jobs/1

I ended up removing the python job because it seemed fast without it and also I think i was only doing a stupid operation in it (merging dataframes) and not any computation? 

**Note**: it looks like many files have changed but that is just the result of merging main into this branch.
The key file being changed (or rather added) is `make_group_file.py`, with a couple of minor changes to `get_genotype_vcf.py` which is just upstream of this.

**Explanation:** in order to test for rare variants ([relevant documentation from SAIGE-QTL](https://weizhou0.github.io/SAIGE-QTL-doc/docs/set_step2.html)), we need an additional input file, which gets added to what we already have:

step 1 (same for common and rare variants; "cv" and "rv"):

- plink files (VRE subset, generated in `get_genotype_vcf.py`)
- pheno_cov filename (phenotype + covariates, generated in `get_anndata.py`)

step 2 (in common):

- VCF with variants to test (generated in `get_genotype_vcf.py`)

For rare variants, we also need a new "group file", vs the "cis window file" used for common variants.

This file specifies which variants should be tested for each gene, and allows to add annotations and weights too.
Here we only add weights for now. Differently from the "cis window file", which only required gene info (chrom, start, end), this requires specifically the variants tested (and their distance from the gene), hence why I am doing this separately (needs a genotype file).


make_group_file.py
```diff
@@ -0,0 +1,189 @@
...
+    Build group file for SAIGE-QTL
+
+    First row: genetic variants to test
+    Second row: annotations (none here)
+    Third row: weights (dTSS here)
+    """
+    data = {'gene': [gene, gene, gene], 'category': ['var', 'anno', 'weight:dTSS']}
+    group_df = pd.DataFrame(data)
+    data = {'var': variants, 'anno': 'null', 'weight:dTSS': weights}
+    vals_df = pd.DataFrame(data).T
```
MattWellie, line 67:
```suggestion
    group_df = pd.DataFrame({'gene': [gene, gene, gene], 'category': ['var', 'anno', 'weight:dTSS']})
    vals_df = pd.DataFrame({'var': variants, 'anno': 'null', 'weight:dTSS': weights}).T
```

make_group_file.py
```diff
@@ -0,0 +1,189 @@
...
+            )
+            variants = [loc.position for loc in ds_result.locus.collect()]
+            gene_tss = int(window_start) + cis_window
+            distances = [int(var) - gene_tss for var in variants]
+            weights = [distance_to_weight(d) for d in distances]
+            group_file_job = get_batch().new_python_job(name=f'group file: {gene}')
+            group_file_job.call(
+                build_group_file_single_gene,
+                gene,
+                str(group_file_path),
```
MattWellie, line 178:
```suggestion
                group_file_path,
```

make_group_file.py
```diff
@@ -0,0 +1,189 @@
...
+import hailtop.batch.job as hb_job
+from typing import List
+
+from cpg_utils import to_path
+from cpg_utils.config import get_config
+
+from cpg_utils.hail_batch import dataset_path, get_batch, init_batch
+
+
+def distance_to_weight(distance: int, gamma: float = 1e-5):
```
MattWellie, line 42: You shouldn't really be importing inside the method definition, which only leaves one functional line in this method. You should probably just do the `math.exp(-gamma * abs(distance))` inside your list comprehension instead of having a method here

helper/variant_counter.py
```diff
@@ -0,0 +1,84 @@
...
+    print(f'Counting variants for {vds_name}')
+    vds = hl.vds.read_vds(vds_path)
+    vds = hl.vds.split_multi(vds, filter_changed_loci=True)
+    # convert to hail matrix table
+    mt = hl.vds.to_dense_mt(vds)
+
+    # filter out loci & variant QC
+    mt = mt.filter_rows(hl.len(mt.alleles) == 2)  # remove hom-ref
+    if exclude_multiallelic:  # biallelic only (exclude multiallelic)
+        print(f'Number of variants before excluding multiallelic loci: {mt.count()[0]}')
```
MattWellie, line 59: bad bad bad bad bad bad badbbadbadbadb

never use a count when you have pending expressions - Each one of these `count()` calls causes the whole result of 

```
vds = hl.vds.read_vds(vds_path)
vds = hl.vds.split_multi(vds, filter_changed_loci=True)
mt = hl.vds.to_dense_mt(vds)
mt = mt.filter_rows(hl.len(mt.alleles) == 2)
```

to be calculated, then thrown away. You're basically tripling the runtime of this script (AFAIK)

helper/variant_counter.py
```diff
@@ -0,0 +1,84 @@
...
+
+    # compute allele frequencies as part of variant qc
+    mt = hl.variant_qc(mt)
+
+    # select common and rare variants
+    cv_mt = mt.filter_rows(mt.variant_qc.AF[1] >= cv_maf_threshold)
+    rv_mt = mt.filter_rows(mt.variant_qc.AF[1] < rv_maf_threshold)
+
+    # count up both donors and variants
+    n_vars, n_donors = cv_mt.count()
```
MattWellie, line 75: here too, the count causes all the calculations to be run, but doesn't save any results. 

On main you should really delete all these count statements. If you really want to log results, you can count the cols/rows before you do any operations, then after all the operations are applied write the result to disk - read that MT/VDS back in, then count the remaining rows. That would do the minimum amount of wasted computation. A freshly read MT will have all the Row and Col counts in the index files, which makes it a free operation

MattWellie (commented):
I've had a scan through here, some I'm not really qualified to look at. I'm not sure the purpose of `helper/variant_counter.py`, is there ever an intention to run this on main? If so I would really review the way you're using logging statements - counting in Hail is not free, so you're doubling/tripling the runtime of these scripts just to print counts

make_group_file.py
annacuomo: ok!

annacuomo:
> I've had a scan through here, some I'm not really qualified to look at. I'm not sure the purpose of `helper/variant_counter.py`, is there ever an intention to run this on main? If so I would really review the way you're using logging statements - counting in Hail is not free, so you're doubling/tripling the runtime of these scripts just to print counts

The variant counter is technically not part of this PR and I have run it in `main` already (once per file) because I needed to get those numbers? 

Happy to review it - is it mostly a matter of adding checkpoints? I do need those numbers for figures / methods but happy to do it more efficiently, although unless I am missing something this wasn't especially expensive? See screenshot
Screenshot 2024-08-21 at 8 37 19 AM

helper/variant_counter.py
annacuomo: haha okok! Is the answer to check point after those 4 lines?

helper/variant_counter.py
annacuomo: ok! so count once at the beginning, do whatever operations i need, save, read back in and count again? I'll give it a go

annacuomo:
> I've had a scan through here, some I'm not really qualified to look at. I'm not sure the purpose of `helper/variant_counter.py`, is there ever an intention to run this on main? If so I would really review the way you're using logging statements - counting in Hail is not free, so you're doubling/tripling the runtime of these scripts just to print counts

Also can you clarify what you don't feel qualified to look at? So I can ask someone else? :)

PR #78, opened 2024-03-07
Adds arguments from config

MattWellie:
Sources all possible arguments from config

For the R script builders, this directly connects the command-string builder to the config contents. This means that the name of the config entry doesn't map 1:1 with what you had before, e.g.

input: `sample_covs`
passed on: `sample_cov_col_list=sample_covs`
into Rscript: `--sampleCovarColList={sample_cov_col_list}`

This value is now just in the config as `sampleCovarColList`


annacuomo (approved):
Ran successfully here: https://batch.hail.populationgenomics.org.au/batches/434728

so looks good to me!! :)

PR #79, opened 2024-03-08
make fill_sex fill_age false and get dummies for seq lib

annacuomo:
I make two main changes here:

- turn `sequencing library` from a categorical covariate to multiple dummy variables (one per sequencing library, containing 1's and 0's reflecting whether a cell is or isn't in that library)
- did the same for cohort (BioHEART / TOB)
- added new covariates in toml file
- make the flags `fill_sex` and `fill_age` -- which fill out numbers for missing values (using mean age for missing age, and 0 i.e. `missing` for missing sex, vs 1 for males and 2 for females) -- False by default

I have not tested the whole script but just the new lines in a notebook and it works as expected:

Screen Shot 2024-03-08 at 5 56 28 pm


get_anndata.py
```diff
@@ -126,8 +126,14 @@ def make_pheno_cov(
...
         mean_age = sample_covs_df['age'].mean()
     cell_ind_df = expression_adata.obs.loc[
-        :, ['cell', 'individual', 'total_counts', 'sequencing_library']
+        :, ['cell', 'individual', 'total_counts', 'cohort', 'sequencing_library']
     ]
+    # make sequencing_library from categorical to dummy numerical covs
+    seq_lib_df = pd.get_dummies(cell_ind_df['sequencing_library'])
+    # do the same for cohort
+    cohort_df = pd.get_dummies(cell_ind_df['cohort'])
+    cell_ind_df = pd.concat([cell_ind_df, cohort_df, seq_lib_df], axis=1)
```
hopedisastro, line 135: Do you still want to leave the original  `sequencing_library` and `cohort` columns in the df, even if you won't use them?

annacuomo: Yes I thought about it but I think they're clearer in their original form, and i always need to specify which covariates I want to test within the saige command so there's no risk they get included accidentally

hopedisastro (approved):

PR #81, opened 2024-03-12
add 5 geno and 5 expression pcs as covs

annacuomo:
add 5 geno and 5 expression PCs as covariate in the toml


hopedisastro:
fyi you can run something in `main` and use a local config file (not committed to the repo).

hopedisastro (approved):

annacuomo:
> fyi you can run something in `main` and use a local config file (not committed to the repo).

yea i know but i wanted to commit the most likely settings just in case i forget everything when i am back lol

PR #82, opened 2024-03-12
convert "-" to "_" in sequencing library covariates

hopedisastro:
Attempting to fix this error: https://batch.hail.populationgenomics.org.au/batches/435305/jobs/1 that results because sequencing library covariates is inputted into an equation model.


annacuomo (approved):
LGTM, thank you!

Issue #74, opened 2024-03-04
saige_assoc.py review


katiedelange:
Some points to consider in saige_assoc.py

- Not all of the SAIGE parameters are self-explanatory, and I can't find explanations for everything in the [docs](https://weizhou0.github.io/SAIGE-QTL-doc/) yet. Could you add some details to the README for anything that we might expect to have to vary (e.g. what is a reasonable value range for `tol` and how might I go about selecting that, or am I never going to want to change it?)

- I assume [this subsetting](https://github.com/populationgenomics/saige-tenk10k/blob/6e556260b457dfc625b47c594becd7751377d8c0/saige_assoc.py#L420) needs to be removed eventually

annacuomo:
> Some points to consider in saige_assoc.py
> 
> * Not all of the SAIGE parameters are self-explanatory, and I can't find explanations for everything in the [docs](https://weizhou0.github.io/SAIGE-QTL-doc/) yet. Could you add some details to the README for anything that we might expect to have to vary (e.g. what is a reasonable value range for `tol` and how might I go about selecting that, or am I never going to want to change it?)
> * I assume [this subsetting](https://github.com/populationgenomics/saige-tenk10k/blob/6e556260b457dfc625b47c594becd7751377d8c0/saige_assoc.py#L420) needs to be removed eventually

This is an excellent point, I'll add some of these in the SAIGE-QTL docs but then also in the readme here! 

Also as a related aside I just realised that most of these are listed but not actually customisable! I added a couple for debugging purposes recently (https://github.com/populationgenomics/saige-tenk10k/blob/real-association/saige_assoc.py#L355-L356), but I think this is not ideal.

I think lint might complain if I add seven million arguments / paraments with click, so maybe I should add some sort of config file instead? It's one of the things on my to do list..

annacuomo:
> I assume [this subsetting](https://github.com/populationgenomics/saige-tenk10k/blob/6e556260b457dfc625b47c594becd7751377d8c0/saige_assoc.py#L420) needs to be removed eventually

@katiedelange yup this is just for testing :)

annacuomo:
@katiedelange back to checking these issues you left a while back! 
I think this one can be closed now as your main points have been addressed? 

[...]

PR #83, opened 2024-03-12
update params description in readme

annacuomo:
Should hopefully be finished now, addressing @katiedelange's main comment from https://github.com/populationgenomics/saige-tenk10k/issues/74.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,8 +108,8 @@ Single-variant association testing ([common variants step 2](https://weizhou0.gi
 * ```min_maf```: minimum minor allele frequency (MAF) (default: 0)
 * ```min_mac```: minimum minor allele count (MAC) (default: 5)
 * ```loco_bool```: boolean specifying whether leave-one-chromosome-out should be used (default: ```FALSE```)
-* ```n_markers```: int = 10000,
-* ```spa_cutoff```: int = 10000,
+* ```n_markers```: internal parameter to batcify variants tested (default: 10000),
+* ```spa_cutoff```: internal parameter to do with the saddlepoint approximation, does not make much of a difference for us (default: 10000),
 
 Obtain gene-level p-values ([common variants only, step 3](https://weizhou0.github.io/SAIGE-QTL-doc/docs/gene_step3.html))
 
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ Script: get_genotype_vcf.py
 Variant selection for VCF files:
 
 * variants that are: i) QC-passing, ii) not ref-ref variants, and iii) not indels or multi-allelic SNPs (when run with --exclude-indels and --exclude-multiallelic).
-* variants that are common (MAF > 0.01) in our population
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -62,9 +70,56 @@ Outputs:
...
+### SAIGE-QTL parameters explanation
+
+Clarifying the reasoning behind the parameters / flags used to run SAIGE-QTL.
+Most of these are (or will be) included in the official [documentation](https://weizhou0.github.io/SAIGE-QTL-doc/).
+
+Fit null model ([step 1](https://weizhou0.github.io/SAIGE-QTL-doc/docs/step1.html)):
+
+* ```pheno_file```: path specifying the location of the phenotype covariate file described above (build during part 2 of the pipeline)
+* ```cov_col_list```: string specifying the columns of the pheno_file that should be used as covariates (separated by a comma, no spaces)
+* ```sample_cov_col_list```: same as above, but specifying only, out of the columns above, the ones that are well defined at individual level (e.g., sex, age, ancestry PCs)
```
katiedelange, line 82: I don't quite get why this distinction is here: when would we use cov_col_list vs sample_cov_col_list?

README.md
```diff
@@ -62,9 +70,56 @@ Outputs:
...
+* ```pheno_col```: specify the column that should be used as phenotype, in our case the gene to test
+* ```trait_type```: specify the model to be used, ```count``` is the Poisson model which should be used here.
+* ```skip_vre```: boolean specifying whether the variance ratio estimation should be run or skipped, should always be false (Note that because the syntax is different between Python and R we encode this as the string ```FALSE``` instead of the boolean ```False```)
+* ```pheno_remove_zeros```: option to remove 0s from phenotype vector (default: ```FALSE``` as it does not make sense for the very sparse scRNA-seq data)
+* ```use_sparse_grm_null```: use sparse GRM to account for relatedness. This is implemented but would require a step0 in the pipeline to first construct this, which is not there at the moment (default: ```FALSE```)
+* ```use_grm_null```: same as above, but without the sparse option (default: ```FALSE```)
+* ```is_cov_offset```: set one of the covariates as offset (default: ```FALSE```)
```
katiedelange, line 92: Which covariate gets set as an offset, and what does this mean/do?

katiedelange (commented):
Mostly looks great, thanks Anna! Just a couple of queries about a few parameters, if you could expand the details for those it would be really helpful.

README.md
annacuomo: sample covs list is always a subset of covs list, just specifying which are well specified at individual level, which allows to treat them more cheaply as the unique values are fewer!

README.md
annacuomo: sorry that is not a good description, this actually is a boolean which adds an offset of ones if there are no covariates -- for us it is always false as we have plenty of covariates. I'll change this!

katiedelange (approved):
Great, thanks for adding those clarifications!

PR #85, opened 2024-03-14
Read covariate file from analysis bucket, not main bucket 

hopedisastro:
FileNotFound error because covariate file is being written to `analysis` bucket (https://github.com/populationgenomics/saige-tenk10k/blob/4513ea0043749a5e43c5131d9d6b1c00e841c378/get_sample_covariates.py#L169), but this get_anndata.py script is expecting the file to be in main. 

Solution is to add `analysis` as a param when reading in the input file.


annacuomo (approved):
Good catch!

PR #86, opened 2024-03-14
Sequencing library notation issue - replace - with _ in config file

hopedisastro:


annacuomo (approved):
Good catch! Can we make a note for Blake to rename not just the TOB "maxi pools" (like `S0021_24a`), but also at least `S0058_59v2` and any other with a hyphen?

hopedisastro:
This is still erroring out: https://batch.hail.populationgenomics.org.au/batches/435978/jobs/91 :/

PR #87, opened 2024-03-14
Optimise memory and storage req for anndata.py

hopedisastro:
Drop the costs by 1/3 by removing explicit storage specs. 

Add `job-cpu` and `job-memory` param in case larger anndata files need more resources. 

Before: https://batch.hail.populationgenomics.org.au/batches/435958
After: https://batch.hail.populationgenomics.org.au/batches/436000


annacuomo (approved):
LGTM, thanks @hopedisastro !

PR #88, opened 2024-03-14
Drop rows with NA values in the sex,age,pc dataframe 

hopedisastro:
NA-containing rows seem to be problematic for SAIGE - https://batch.hail.populationgenomics.org.au/batches/435978/jobs/91, so we specify if `fill_in_sex` or `fill_in_age` is False, ensure that rows with missing values are dropped prior to making the pheno_cov files.

### DIFF
--- a/get_anndata.py
+++ b/get_anndata.py
@@ -144,7 +144,6 @@ def make_pheno_cov(
         sample_covs_cells_df['sex'] = sample_covs_cells_df['sex'].fillna(0)
     if fill_in_age:
         sample_covs_cells_df['age'] = sample_covs_cells_df['age'].fillna(mean_age)
-    if fill_in_sex is False or fill_in_age is False:
         # drop rows with missing values (SAIGE throws an error otherwise:  https://batch.hail.populationgenomics.org.au/batches/435978/jobs/91)
         sample_covs_cells_df = sample_covs_cells_df.dropna()
     gene_adata = expression_adata[:, expression_adata.var['gene_name'] == gene]
--- a/get_anndata.py
+++ b/get_anndata.py
@@ -144,6 +144,9 @@ def make_pheno_cov(
         sample_covs_cells_df['sex'] = sample_covs_cells_df['sex'].fillna(0)
     if fill_in_age:
         sample_covs_cells_df['age'] = sample_covs_cells_df['age'].fillna(mean_age)
+    if fill_in_sex is False or fill_in_age is False:
+        # drop rows with missing values (SAIGE throws an error:  https://batch.hail.populationgenomics.org.au/batches/435978/jobs/91)
+        sample_covs_cells_df = sample_covs_cells_df.dropna()
     gene_adata = expression_adata[:, expression_adata.var['gene_name'] == gene]
     gene_name = gene.replace("-", "_")
     expr_df = pd.DataFrame(
--- a/get_anndata.py
+++ b/get_anndata.py
@@ -145,7 +145,7 @@ def make_pheno_cov(
     if fill_in_age:
         sample_covs_cells_df['age'] = sample_covs_cells_df['age'].fillna(mean_age)
     if fill_in_sex is False or fill_in_age is False:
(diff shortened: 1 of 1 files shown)


get_anndata.py
```diff
@@ -144,6 +144,9 @@ def make_pheno_cov(
         sample_covs_cells_df['sex'] = sample_covs_cells_df['sex'].fillna(0)
     if fill_in_age:
         sample_covs_cells_df['age'] = sample_covs_cells_df['age'].fillna(mean_age)
+    if fill_in_sex is False or fill_in_age is False:
+        # drop rows with missing values (SAIGE throws an error:  https://batch.hail.populationgenomics.org.au/batches/435978/jobs/91)
```
hopedisastro, line 148:
```suggestion
        # drop rows with missing values (SAIGE throws an error otherwise:  https://batch.hail.populationgenomics.org.au/batches/435978/jobs/91)
```

annacuomo:
I think this doesn't cover the case where `fill_in_sex` and `fill_in_age` are set to True, so there are no NAs there, but there still are NAs in the genotype PCs? Probably ok for now since those by default are False, so NAs would be dropped, but a better solution would be to add an option to fill in missing genotype PCs as well like we do with age (using the mean), and then adding that as an option that needs.

annacuomo:
Or actually what about dropping NAs regardless? if the filler flags are True there won't be any and we will just not drop anything at all

get_anndata.py
```diff
@@ -144,6 +144,9 @@ def make_pheno_cov(
         sample_covs_cells_df['sex'] = sample_covs_cells_df['sex'].fillna(0)
     if fill_in_age:
         sample_covs_cells_df['age'] = sample_covs_cells_df['age'].fillna(mean_age)
+    if fill_in_sex is False or fill_in_age is False:
```
hopedisastro, line 147:
```suggestion
```

get_anndata.py
```diff
@@ -144,6 +144,9 @@ def make_pheno_cov(
...
+        # drop rows with missing values (SAIGE throws an error otherwise:  https://batch.hail.populationgenomics.org.au/batches/435978/jobs/91)
+        sample_covs_cells_df = sample_covs_cells_df.dropna()
```
hopedisastro, line 149:
```suggestion
sample_covs_cells_df = sample_covs_cells_df.dropna()
```

get_anndata.py
```diff
@@ -144,6 +144,9 @@ def make_pheno_cov(
         sample_covs_cells_df['sex'] = sample_covs_cells_df['sex'].fillna(0)
     if fill_in_age:
         sample_covs_cells_df['age'] = sample_covs_cells_df['age'].fillna(mean_age)
+    if fill_in_sex is False or fill_in_age is False:
+        # drop rows with missing values (SAIGE throws an error otherwise:  https://batch.hail.populationgenomics.org.au/batches/435978/jobs/91)
```
hopedisastro, line 148:
```suggestion
# drop rows with missing values (SAIGE throws an error otherwise:  https://batch.hail.populationgenomics.org.au/batches/435978/jobs/91)
```

get_anndata.py
```diff
@@ -144,6 +144,8 @@ def make_pheno_cov(
         sample_covs_cells_df['sex'] = sample_covs_cells_df['sex'].fillna(0)
     if fill_in_age:
         sample_covs_cells_df['age'] = sample_covs_cells_df['age'].fillna(mean_age)
+# drop rows with missing values (SAIGE throws an error otherwise:  https://batch.hail.populationgenomics.org.au/batches/435978/jobs/91)
```
hopedisastro, line 147:
```suggestion
   # drop rows with missing values (SAIGE throws an error otherwise:  https://batch.hail.populationgenomics.org.au/batches/435978/jobs/91)
```

hopedisastro:
@annacuomo

annacuomo (approved):
Great, looks good to me now :)

PR #90, opened 2024-03-19
Change default min_pct to 1 from 5

hopedisastro:
The default is set to 5, but I checked with Anna and she had decided to set it to 1(%), so this PR reflects that change.


michael-harper (approved):
Looks good, just wanting to point out that in `filter_lowly_expressed_genes()` the default for`min_pct` was already being set to 1. This shouldn't have effected any runs as it would've been passed 5 anyway given we set the default earlier but just a heads up

PR #91, opened 2024-04-27
Replace VDS version param with VDS path for flexibility AND fix bug in re-partitioning of the MT checkpoints

hopedisastro:
The current implementation allows users to specify the VDS version only (assumes that a joint TOB+BioHEART analysis will be performed). 

Here I amend the param to specify the VDS path, allowing more flexibility to specify which VDS to be used (TOB-only, BioHEART-only, or combined VDS). 

Also debugged the re-partitioning of the checkpointed matrix tables in `checkpoint_mt()`

### DIFF
--- a/get_anndata.py
+++ b/get_anndata.py
@@ -245,7 +245,6 @@ def main(
     )
 
     for celltype in celltypes.split(','):
-
         # extract cell-level covariates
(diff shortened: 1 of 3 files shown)


MattWellie (approved):
molto bene

PR #93, opened 2024-04-30
More flexible params for get anndata, removal of dummy variables for `sequencing_library` and `cohort`

hopedisastro:
Changes: 
1) More flexible params for the script:
- specifying the GCP paths for the cell type and sample level covariate files as well as for the anndata files. 

2) Remove:
- `sequencing_library` dummy variables because they will no longer be included in the model - expensive to run for freeze 1, possible consider for freeze 2; cost scales pretty much linearly with every additional covariate we add to the model. 
- `cohort` dummy variable because cohort-specific analysis will be conducted
(these pertain to the make_pheno_cov() helper method) 

3) Use `subprocess` to copy the filtered h5ad file to GCP instead of the existing hl.hadoop read/write because the latter produced error - https://batch.hail.populationgenomics.org.au/batches/444912/jobs/1

4) Do not loop over `gene_name` in the H5AD file because some genes with distinct ENSG IDs have the same `gene_name` entry so the job will fail. Instead, we now loop over expression_adata.var.index. See failed jobs here -https://batch.hail.populationgenomics.org.au/batches/445040?q=state%3Dfailed

fyi @annacuomo 

TODO: When scRNA freeze 2 comes, we need to delete the code that amends the cell barcode suffix. Cell barcodes dont' currently match between H5AD file and the cell covariate file because one of them has the sequencing library attached as as suffix and the other doesn't (Blake has been informed)

### DIFF
--- a/get_anndata.py
+++ b/get_anndata.py
@@ -87,7 +87,7 @@ def get_gene_cis_info(
     gene_info_df = copy_h5ad_local_and_open(gene_info_df_path).var
 
     # select the gene from df
-    gene_info_gene = gene_info_df[gene_info_df['gene_name'] == gene]
+    gene_info_gene = gene_info_df[gene_info_df.index == gene]
     # get gene chromosome
     chrom = gene_info_gene['chr'][0]
     # remove "chr"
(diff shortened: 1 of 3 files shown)


hopedisastro:
https://batch.hail.populationgenomics.org.au/batches/445031/jobs/1 error when you specify more than one chromosome

michael-harper (approved):
Hard to visualise dataframe structures from this, but LGTM

PR #99, opened 2024-05-09
add relateds to drop ht as a dynamic file path

hopedisastro:
Easier as a dynamic file path because the subdirectories can easily change; and we know there will at least be two different file paths - one for tob and one for bioheart.

### DIFF
--- a/get_genotype_vcf.py
+++ b/get_genotype_vcf.py
@@ -33,6 +33,15 @@ analysis-runner \
     --output-dir "saige-qtl/input_files/genotypes/" \
     python3 get_genotype_vcf.py --vds-path=gs:// --chromosomes chr1,chr2,chr22
 
+
+    analysis-runner \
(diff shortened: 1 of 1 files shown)


get_genotype_vcf.py
```diff
@@ -204,6 +213,10 @@ def remove_chr_from_bim(input_bim: str, output_bim: str, bim_renamed: str):
 @click.option('--exclude-multiallelic', is_flag=False)
 @click.option('--exclude-indels', is_flag=False)
 @click.option('--plink-job-storage', default='1G')
+@click.option(
+    '--relateds-to-drop-path',
+    default='gs://cpg-bioheart-test/large_cohort/v1-0/relateds_to_drop.ht',
```
michael-harper, line 218: Does it make sense to have a default value here?

michael-harper (approved):

PR #100, opened 2024-05-12
Add sentinels (fake jobs) so that we can do `always_run()` but also cancel the jobs if needed. 

hopedisastro:
I think this is right....? 

Added sentinel jobs to each of the four steps: 
1. Fitting the null model `run_fit_null_job`
2. ` build_run_single_variant_test_command`
3. `build_obtain_gene_level_pvals_command`
4. `summarise_cv_results` (sentinel added in `main()`) 

Ensured that the sentinel job was the one being returned in steps 1,2,3 (ie we want to add the sentinel jobs to `manage_concurrency_for_job`)

### DIFF
--- a/saige_assoc.py
+++ b/saige_assoc.py
@@ -328,6 +328,7 @@ def main(
     # pull principal args from config
     chromosomes: list[str] = get_config()['saige']['chromosomes']
     celltypes: list[str] = get_config()['saige']['celltypes']
+    celltype_jobs: dict[str, list] = dict()
 
     vre_plink_path = f'{vre_files_prefix}/vre_plink_2000_variants'
 
@@ -421,6 +422,10 @@ def main(
                     job3.depends_on(gene_dependency)
                     gene_dependency = job3
 
+                # add this job to the list of jobs for this cell type
+                if gene_dependency is not None:
+                    celltype_jobs.setdefault(celltype, []).append(gene_dependency)
+
     # summarise results (per cell type)
     for celltype in celltypes:
         logging.info(f'start summarising results for {celltype}')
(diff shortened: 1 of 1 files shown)


MattWellie:
I don't think that's quite how I pictured this - you have a loop over a number of genes, and for each gene you have a chain of jobs which have an internal dependency. 

In that scenario you need a single fake job per gene to make sure that the gene begins processing, then create a dependency chain from there so that the jobs run, but dependent on the success of the previous job. 

This change creates an un-killable fake job for each each task within each gene, even though there are implicit dependencies from using the output of the previous job. If a job fails, tasks dependent on it may still attempt to start due to the always-run dependency.

My suggestion:
```
for gene in ...:
    gene_dependency = <create a fake job, always_run>
    job1, job_1_output = run_fit_null_job(...)
    if job1 is not None:
          job1.depends_on(gene_dependency)
          gene_dependency = job1
    ....
```

This means:

- each gene is guaranteed to start, not that each individual job for a gene is guaranteed to start
- each job can return None (if no job is required), or a job
- the dependency is rotated, so that if a real job is added for a gene, the next depends on that real job (and its successful completion)

saige_assoc.py
```diff
@@ -419,11 +432,16 @@ def manage_concurrency_for_job(job: hb.batch.job.Job):
         summary_output_path = (
             f'output_files/summary_stats/{celltype}_all_cis_cv_results.tsv'
         )
-
+        summarise_fake_job = get_batch().new_job(
+            f'Fake summarise CV results for {celltype}'
+        )
+        summarise_fake_job.depends_on(*jobs)
+        summarise_fake_job.always_run()
+        summarise_fake_job.command('echo "fake summarise CV results"')
```
MattWellie, line 440: I don't think there should be an always_run summary job - making sure each independent gene's chain of jobs starts is one thing, but this job would always run, and would be 'summarising' an incomplete batch. I think it's accurate that this should only run if all jobs up to this point have succeeded?

MattWellie (approved):

PR #102, opened 2024-07-01
add shuffled sample ids

annacuomo:
In order to run calibration analysis (to assess whether in the absence of any true signal the association method is not inflated) we want to run SAIGE-QTL with permuted sample IDs. Here, I create 10 shuffled versions of sample ID. Note that the permutation should be between individuals, not cells so this is a good place to do this.

### DIFF
--- a/get_sample_covariates.py
+++ b/get_sample_covariates.py
@@ -176,7 +176,9 @@ def main(
         )
 
     # save to file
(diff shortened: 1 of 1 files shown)


MattWellie (approved):
so you're adding `n` new columns, each containing a new ordering of the sample IDs? Seems fine.

annacuomo:
> so you're adding `n` new columns, each containing a new ordering of the sample IDs? Seems fine.

yup that's it, thank you!

PR #106, opened 2024-08-15
Add sequencing library and cohort back as covariates

hopedisastro:
Was removed before because wasn't planning to use them as covariates.


annacuomo:
needs blacking but otherwise looks good to me!
Good to include cohort as well for possible joint analysis :)

annacuomo (approved):

PR #112, opened 2024-08-26
fresh group file making script

annacuomo:
Fresh version of https://github.com/populationgenomics/saige-tenk10k/pull/68 to avoid confusion about what is to be reviewed here.

Successfully ran in `test` here: https://batch.hail.populationgenomics.org.au/batches/479045/jobs/1

**Copy of explanation:** in order to test for rare variants ([relevant documentation from SAIGE-QTL](https://weizhou0.github.io/SAIGE-QTL-doc/docs/set_step2.html)), we need an additional input file, which gets added to what we already have:

step 1 (same for common and rare variants; "cv" and "rv"):

plink files (VRE subset, generated in get_genotype_vcf.py)
pheno_cov filename (phenotype + covariates, generated in get_anndata.py)
step 2 (in common):

VCF with variants to test (generated in get_genotype_vcf.py)
For rare variants, we also need a new "group file", vs the "cis window file" used for common variants.

This file specifies which variants should be tested for each gene, and allows to add annotations and weights too.
Here we only add weights for now. Differently from the "cis window file", which only required gene info (chrom, start, end), this requires specifically the variants tested (and their distance from the gene), hence why I am doing this separately (needs a genotype file).

@silkm @michael-harper, Katie (@katiedelange ) suggested adding you as reviewers, but let me know if you think support from someone from the software team is required also.

### DIFF
--- a/make_group_file.py
+++ b/make_group_file.py
@@ -59,12 +59,6 @@ def main(
     for chrom in chromosomes.split(','):
         print(f'chrom: {chrom}')
 
-        # load rare variant vcf file for specific chromosome
-        vcf_path = dataset_path(
-            f'saige-qtl/bioheart/input_files/genotypes/vds-bioheart1-0/{chrom}_rare_variants.vcf.bgz'
-        )
-        ds = hl.import_vcf(vcf_path, reference_genome='GRCh37')
-
         # do a glob, then pull out all file names as Strings
         files = [
             str(file)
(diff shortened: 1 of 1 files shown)


silkm:
I think looks good. Optional thought - would it be easier to have your `files[0:10]` as a click argument for number of files to test up to so you can rerun without needing to PR to change how many to run on?

silkm (approved):

silkm:
I've approved, up to you if you want to wait for @michael-harper to have a look or give thoughts on if you need software to take a look also

annacuomo:
> I think looks good. Optional thought - would it be easier to have your `files[0:10]` as a click argument for number of files to test up to so you can rerun without needing to PR to change how many to run on?

good idea! I added it as a click argument!

ckrishk (approved):
LGTM

make_group_file.py
```diff
@@ -0,0 +1,129 @@
...
+            )
+            for f in files
+        ]
+        logging.info(f'I found these genes: {", ".join(genes)}')
+
+        # load rare variant vcf file for specific chromosome
+        vcf_path = dataset_path(
+            f'saige-qtl/bioheart/input_files/genotypes/{vds_name}/{chrom}_rare_variants.vcf.bgz'
+        )
+        ds = hl.import_vcf(vcf_path, reference_genome='GRCh37')
```
ckrishk, line 89: nit: may be a good idea to keep this parameter configurable

ckrishk (approved):

make_group_file.py
annacuomo: done!

PR #111, opened 2024-08-26
Donor counter

annacuomo:
SAIGE-QTL should run fine when the genotype files in step 1 (VRE plink files) and step 2 (VCF with actual variants to test) contain the same donors.

Instead, it is failing at step 2 because of donor mismatch: https://batch.hail.populationgenomics.org.au/batches/478919 (_At least one subject requested is not in VCF file._)

So I created this helper function to check the donors match across files.

Ran successfully in `test` and should be pretty quick / cheap: https://batch.hail.populationgenomics.org.au/batches/479010/jobs/1

PR #113, opened 2024-08-26
remove related before making vre files also

annacuomo:
I noticed that there was a donor mismatch between the genotype files included in steps 1 & 2 of the SAIGE-QTL pipeline, and confirmed by counting the donors in the different files, see PR: https://github.com/populationgenomics/saige-tenk10k/pull/111 and successful run in `main` confirming mismatching numbers:  https://batch.hail.populationgenomics.org.au/batches/479015/jobs/1

I think the mismatch stems from a recent improvement where we remove related samples from the vds / mts which we export as VCF files, but I forgot to do so for the vre plink files as well, so rectifying this here.

Ran successfully in `test`: https://batch.hail.populationgenomics.org.au/batches/479073/jobs/1

And donor counter to confirm matching number of donors: https://batch.hail.populationgenomics.org.au/batches/479080/jobs/1 (there are only 5 donors in the `test` object so this will have to be further checked in `main`)

### DIFF
--- a/get_genotype_vcf.py
+++ b/get_genotype_vcf.py
@@ -22,7 +22,8 @@ analysis-runner \
     --dataset "bioheart" \
     --access-level "test" \
     --output-dir "saige-qtl/bioheart/input_files/genotypes" \
-    python3 get_genotype_vcf.py --vds-path=gs://cpg-bioheart-test/vds/bioheart1-0.vds --chromosomes chr1,chr2,chr22 --vre-mac-threshold 1
+    python3 get_genotype_vcf.py --vds-path=gs://cpg-bioheart-test/vds/bioheart1-0.vds \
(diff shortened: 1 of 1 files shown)


michael-harper (approved):
LGTM. Quite a long script `main` function though

annacuomo:
> LGTM. Quite a long script `main` function though

Agreed! The bit removing related samples will all go as soon as we get the final VDS from the LC pipeline, which will have performed sample QC already! So quite a few lines will go then :)

annacuomo:
Just to confirm that after removing the related samples from the VRE plink files (this PR), re-running `get_genotype_vcf.py` (https://batch.hail.populationgenomics.org.au/batches/479096/jobs/1), and then re-running the donor counter (https://batch.hail.populationgenomics.org.au/batches/479128/jobs/1), the donor mismatch is resolved :)

PR #114, opened 2024-08-28
add low frequency variants

annacuomo:
Add ability to count variants near the common frequency threshold as well.

Noting that Matt commented on this in an old PR (https://github.com/populationgenomics/saige-tenk10k/pull/68), highlighting how more checkpoints should be used I think, to avoid duplicated computation? Any help there would be useful, but also as I noted in that PR these operations are very cheap.

### DIFF
--- a/helper/variant_counter.py
+++ b/helper/variant_counter.py
@@ -78,7 +78,7 @@ def count_variants(
 
     # count up both donors and variants
     n_common_vars, n_donors = cv_mt.count()
-    n_low_frequency_vars, n_donors = lf_mt.count()[0]
+    n_low_frequency_vars = lf_mt.count()[0]
     n_rare_vars = rv_mt.count()[0]
 
     print(f'Donor count: {n_donors}')
--- a/helper/variant_counter.py
+++ b/helper/variant_counter.py
@@ -28,7 +28,7 @@ from cpg_utils.hail_batch import init_batch
 @click.command()
 @click.option('--vds-path', required=True)
 @click.option('--cv-maf-threshold', default=0.05)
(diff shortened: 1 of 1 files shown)


MattWellie (approved):
In general count operations should be reduced/removed, as they're computationally very expensive and none of the processing done is saved. This script is just counting with no saved output, so I don't know what purpose it serves, but this change is fine I guess 😂

annacuomo:
> In general count operations should be reduced/removed, as they're computationally very expensive and none of the processing done is saved. This script is just counting with no saved output, so I don't know what purpose it serves, but this change is fine I guess 😂

The purpose this serves is that I need to know these numbers to write into the paper / know them when people ask when I present these data. Specifically in this case (for the TOB cohort) we also have SNP array data so I wanted to know how many more variants we can call for WGS to make plots / write ups. The SNP data is just plink files on HPC so I can instantly get these numbers, here everything is always harder. If there is another obvious way to do this, e.g. save a summary table with AF info into the `analysis` bucket, or at least make this code more efficient, I am happy to hear suggestions.

MattWellie:
> The purpose this serves is that I need to know these numbers to write into the paper

I thought it would be, in which case this is fine. As a general rule, doing count operations just for info is wasteful as it doubles the amount of calculation being performed. Here the counting is the sole purpose, so it's unavoidable. 

You're not doing many filters/operations on the data so this shouldn't be too bad, and using this kind of script is preferable to changing other scripts that already work just to include additional counts. 

My improvement suggestion would be store the results in a file instead of just logging the lines inside the hail batch job. It would be more robust to export the same data in hard copy so that you don't need to rely on finding the same batch and copy/pasting to keep the results.

PR #116, opened 2024-08-29
add variant ids to vcf based on other cols

annacuomo:
The pipeline finally ran successfully, but I noticed the final summary file `cpg-bioheart-main-analysis/saige-qtl/bioheart_n990_and_tob_n1055/output_files/v3/output_files/summary_stats/B_naive_all_cis_cv_results.tsv` has `.` for `top_MarkerID` because the `ID` column of the VCFs I am testing is empty.

So going back to the script generating these VCF files and annotating them using bcftools so that the variant ID includes info on chrom, pos, ref and alt alleles in the form `{chrom}_{pos}_{ref}_{alt}`.

### DIFF
--- a/get_genotype_vcf.py
+++ b/get_genotype_vcf.py
@@ -162,7 +162,7 @@ def add_remove_chr_and_index_job(vcf_path):
     )
     bcftools_job.command(
         f"""
-        bcftools annotate --rename-chrs chr_update.txt {vcf_input} | \\
+        bcftools annotate --rename-chrs chr_update.txt --set-id +'%CHROM\_%POS\_%REF\_%FIRST_ALT' {vcf_input} | \\
         bgzip -c > {bcftools_job.output['vcf.bgz']}
         bcftools index -c {bcftools_job.output['vcf.bgz']}
     """


get_genotype_vcf.py
```diff
@@ -162,7 +162,7 @@ def add_remove_chr_and_index_job(vcf_path):
     )
     bcftools_job.command(
         f"""
-        bcftools annotate --rename-chrs chr_update.txt {vcf_input} | \\
+        bcftools annotate --rename-chrs chr_update.txt --set-id +'%CHROM\_%POS\_%REF\_%FIRST_ALT' {vcf_input} | \\
```
michael-harper, line 165: Just flagging that the documentation says that the `+` at the start of the `--set-id` command then only missing IDs will be set. So if not all IDs are missing then there will be different formatting for the ID field in the vcf file

michael-harper (approved):

get_genotype_vcf.py
annacuomo: good flag! I think the way Hail exports to vcf it would only add an ID if there were [`rsids`](https://www.pharmgkb.org/page/glossary#:~:text=rsID%20number,to%20a%20specific%20genetic%20variant.) in the VDS / MT object, which we specifically are avoiding doing, because they can be ambiguous compared to just specifying the chromosome, position and reference / alternative allele. So it will always be missing for us, or at least for the cohorts included in this project.

annacuomo:
Had tested locally already but ran in `test` as well just in case: https://batch.hail.populationgenomics.org.au/batches/480935/jobs/1 (using chr20 so should hopefully not re-run the VRE file bit and go straight to the VCFs).

Screenshot below with correct IDs

Screenshot 2024-09-02 at 5 08 38 PM

PR #117, opened 2024-08-30
start adding to set-based test instead

annacuomo:
Ready for review but not tested in `test` (see below, [this PR](https://github.com/populationgenomics/transfer-private/pull/76), and [this slack message](https://centrepopgen.slack.com/archives/C018KFBCR1C/p1725421880287999)).

Now that group files can be generated as well ([PR](https://github.com/populationgenomics/saige-tenk10k/pull/112)), we are ready to run set-based tests using SAIGE-QTL ([docs](https://weizhou0.github.io/SAIGE-QTL-doc/docs/set_step2.html)).

This PR adds a new file (`saige_assoc_set_test.py`) which is similar to `saige_assoc.py` but uses a different function from SAIGE-QTL. Step 1 (fit the null model) will be identical, but step 2 is a [set-based test](https://weizhou0.github.io/SAIGE-QTL-doc/docs/set_step2.html) and not a single-variant test, and there is no step 3, as the results are already at gene-level. The summarising final step is the same again.

An important note is that if both single-variant and set-based tests are to be performed on the same data, we should make sure step 1 is only run once (which should be achieved by using the same output dir).

----

I cannot find a time when the OG SAIGE-QTL pipeline (common variant, single-variant test) worked in `test`, so trying again here (changing the running instructions in the files below as part of this effort) for chr21, B_naive.

* get genotypes: https://batch.hail.populationgenomics.org.au/batches/482719/jobs/1 🎉 (files in `cpg-bioheart-test/saige-qtl/bioheart_n990_and_tob_n1055/input_files/genotypes/v3/vds-tenk10k1-0`)
* get sample covariates:   https://batch.hail.populationgenomics.org.au/batches/482692/jobs/1 🎉 (file in `cpg-bioheart-test-analysis/saige-qtl/input_files/covariates/sex_age_geno_pcs_shuffled_ids_tob_bioheart.csv`)
* get anndata: https://batch.hail.populationgenomics.org.au/batches/482736/jobs/1 🎉 (pheno files in `cpg-bioheart-test/saige-qtl/bioheart_n990_and_tob_n1055/input_files/pheno_cov_files/B_naive/chr21`, and cis window files in `cpg-bioheart-test/saige-qtl/bioheart_n990_and_tob_n1055/input_files/cis_window_files/chr21`)
* saige assoc (steps 1, 2a, 3): https://batch.hail.populationgenomics.org.au/batches/482741 -- this fails, maybe because there are missing values in the covariates?
* saige assoc try again: https://batch.hail.populationgenomics.org.au/batches/482934 -- trying again with only sex (age had some NAs and genotype PCs were not there at all). Noting that this might still fail because the 5 donors in `test` are too few for SAIGE-QTL..
* saige assoc try again again: https://batch.hail.populationgenomics.org.au/batches/482936 (no covs at all, which I am not even sure is allowed) - failing again, Wei thinks it's just too few donors.

Update on this (again see [thread](https://centrepopgen.slack.com/archives/C018KFBCR1C/p1725421880287999)) is that at the moment with the current genotype files the SAIGE-QTL pipeline cannot be run in `test`, but since it runs successfully in `main` (last successful run: https://batch.hail.populationgenomics.org.au/batches/482945 where only one gene failed and then successfully finished when i re-run with higher convergence iterations so running for that gene and successfully summarising all: https://batch.hail.populationgenomics.org.au/batches/482970) and this new script is very similar to it hopefully this will run also!


silkm (approved):
From what @michael-harper and I have gone over this looks good to us.

I assume the way the R scripts are included was already discussed but just double checking - they're called from `/usr/local/bin` and not tracked as part of this repository, does this pose any issues such as version control / script provenance?

annacuomo:
> From what @michael-harper and I have gone over this looks good to us.
> 
> I assume the way the R scripts are included was already discussed but just double checking - they're called from `/usr/local/bin` and not tracked as part of this repository, does this pose any issues such as version control / script provenance?

Thanks both of you! 

Re: the R scripts, that is an excellent point and yes potentially problematic in terms of versioning. 
To recap, this the [SAIGE-QTL image](https://github.com/populationgenomics/images/blob/main/images/saige-qtl/Dockerfile) we use, so the R scripts are copied over to `/usr/local/bin` from the repo upon installing the tool from GitHub.
Because the tool is still in the development stage, to try and keep track of what version of the tool we used we specified the commit in the [images.toml](https://github.com/populationgenomics/images/blob/main/images.toml#L34). I am hoping when using the specific image we can trace back what version we used but not 100% sure that's true?

annacuomo:
I've created an issue re: SAIGE-QTL versioning! https://github.com/populationgenomics/saige-tenk10k/issues/121

PR #119, opened 2024-09-05
Updating output files path for eQTL analysis

annacuomo:
Results from the SAIGE-QTL pipeline ended up in an awkwardly called path with repeated `output_files` directory, e.g., `gs://cpg-bioheart-main/saige-qtl/bioheart_n990_and_tob_n1055/output_files/v3/output_files/`, so fixing that here.
While I am at it, also split results by chromosome and cell type to make them easier to navigate. 

Because of the too-small genotype files in `test` cannot test this but hoping it is a small enough change that it is not a problem!

### DIFF
--- a/saige_assoc.py
+++ b/saige_assoc.py
@@ -386,7 +386,7 @@ def main(
 
                 # check if these outputs already exist, if so don't make a new job
                 null_job, null_output = run_fit_null_job(
-                    output_path(f'output_files/{celltype}_{gene}'),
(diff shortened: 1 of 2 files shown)


hopedisastro (approved):

annacuomo:
running to test new paths: https://batch.hail.populationgenomics.org.au/batches/483775
https://batch.hail.populationgenomics.org.au/batches/483776

PR #123, opened 2024-09-06
group file should be tab separated

annacuomo:
I think the reason [this](https://batch.hail.populationgenomics.org.au/batches/483004/jobs/2) fails may be that group files should be tab separated, not comma.

See an [example group file](https://github.com/weizhou0/qtl/blob/main/extdata/input/groupFile_temp.txt) in the SAIGE-QTL repo.

On the plus side, step 1 was not re-run (as hoping) and step 2 does at least start, so the [dollar sign removal](https://github.com/populationgenomics/saige-tenk10k/pull/122) was useful!

### DIFF
--- a/make_group_file.py
+++ b/make_group_file.py
@@ -121,7 +121,7 @@ def main(
             group_vals_df = pd.merge(group_df, vals_df, on='category')
             group_file = f'{group_files_path}{chrom}/{gene}_{cis_window}bp.tsv'
             with to_path(group_file).open('w') as gdf:
-                group_vals_df.to_csv(gdf, index=False, header=False)
+                group_vals_df.to_csv(gdf, index=False, header=False, sep='\t')
 
 
 if __name__ == '__main__':


silkm (approved):
I had some trouble confirming if this is the source of your Hail error, as I think your group file is then passed to your R script and I'm not sure what it needs the group file to look like. Approving though as no issue changing to tab separation, hope it fixes it!

annacuomo:
> I had some trouble confirming if this is the source of your Hail error, as I think your group file is then passed to your R script and I'm not sure what it needs the group file to look like. Approving though as no issue changing to tab separation, hope it fixes it!

Thanks for approving!

My guess is that [these lines](https://github.com/weizhou0/qtl/blob/main/R/SAIGE_SPATest_Region_Func.R#L94-L97) are responsible! It is checking that each line has at least three terms and not getting that? So I figured if it's reading the whole line with the commas as a single term that would explain it? And also as you say it doesn't hurt + I shouldn't have a file that is called `.tsv` but is really comma separated 😅

PR #124, opened 2024-09-09
Modify SAIGE-QTL RV flags / args for debugging

annacuomo:
The rare variant set-test run is costing a lot more than expected: https://batch.hail.populationgenomics.org.au/batches/483008 😬 

A couple of changes to debug:
* noticing that `GT` was specified as VCF field for single-variant tests, vs `DS`
* added an argument to specify how many genes are tested `--ngenes-to-test`, so can be run for just one or two genes when debugging
* because the stalling happened when specifying the weights, add option to specify different group file names (e.g. `_no_weights` as generated here: https://github.com/populationgenomics/saige-tenk10k/pull/125, but in the future specify annotations etc)

[Slack thread](https://centrepopgen.slack.com/archives/C030X7WGFCL/p1725919664235749) for some context.

### DIFF
--- a/saige_assoc_set_test.py
+++ b/saige_assoc_set_test.py
@@ -322,6 +322,7 @@ def summarise_cv_results(
     '--vre-files-prefix', default=dataset_path('saige-qtl/input_files/genotypes')
 )
 @click.option('--ngenes-to-test', default='all')
+@click.option('--group-file-specs', default='')
 @click.command()
 def main(
     # output from get_anndata.py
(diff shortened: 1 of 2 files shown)


michael-harper (approved):
Happy with just using the first `ngenes_to_test` files or do you want to be more specific with which ones to use?

annacuomo:
> Happy with just using the first `ngenes_to_test` files or do you want to be more specific with which ones to use?

Good point! I'll leave as is for now to be consistent with what I've done in the past (e.g. https://github.com/populationgenomics/saige-tenk10k/blob/main/make_group_file.py#L75-L77) but will make a note to see if it might make sense to select specific example genes for testing, thank you!

PR #125, opened 2024-09-10
make group file without weights

annacuomo:
In an effort to debug the run described in [this recent post](https://centrepopgen.slack.com/archives/C030X7WGFCL/p1725919664235749), trying to make a different `group file` without weight info. The log ended on these lines (after for some reason printing every single weight 🙄 )

> type  weight:dTSS 
> weightname
> [1] "dTSS

so possibly removing this specific test may help things. More info on group files (and how the `weight` line is not necessary) in the [SAIGE-QTL docs](https://weizhou0.github.io/SAIGE-QTL-doc/docs/set_step2.html#input-files).

Successfully ran in `test`: https://batch.hail.populationgenomics.org.au/batches/483266/jobs/1 🎉 (example newly created group file without weights: `cpg-bioheart-test/saige-qtl/input_files/group_files/chr22/ACO2_100000bp_no_weights.tsv`)

PR #126, opened 2024-09-10
weights as option in group file

annacuomo:
Updated version of https://github.com/populationgenomics/saige-tenk10k/pull/125 to keep more general: have `gamma` as an arg such that we can provide a value if we want to include dTSS weights, or `none` if not.

Ran successfully when 
* including weights (default `gamma`): https://batch.hail.populationgenomics.org.au/batches/483314/jobs/1 (files here: `cpg-bioheart-test/saige-qtl/input_files/group_files/v2/chr22`)
* excluding weights (setting `gamma` = `none`): https://batch.hail.populationgenomics.org.au/batches/483312/jobs/1 (files also here: `cpg-bioheart-test/saige-qtl/input_files/group_files/v2/chr22`, this time with `no_weights` in the name)

Additionally, updating variant definition to match format from the [SAIGE-QTL example group file](https://github.com/weizhou0/qtl/blob/main/extdata/input/nfam_500_nindep_5000_01_group.txt), i.e. `chrom.pos.ref.alt`

### DIFF
--- a/make_group_file.py
+++ b/make_group_file.py
@@ -44,7 +44,7 @@ from cpg_utils.hail_batch import init_batch
 @click.option('--group-files-path')
 @click.option('--vcf-path')
 @click.option('--cis-window', default=100000)
-@click.option('--gamma', default=1e-5)
+@click.option('--gamma', default='1e-5')
 @click.option('--ngenes-to-test', default='all')
 @click.option('--genome-reference', default='GRCh37')
 def main(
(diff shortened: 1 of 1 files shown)


Alexander-Stuckey (approved):
Looks good to me

annacuomo:
running in main: https://batch.hail.populationgenomics.org.au/batches/483339/jobs/1

PR #127, opened 2024-09-11
space separated instead of tab

annacuomo:
Very conscious that this all feels like trial and error without much reasoning behind it, but Wei suggested saving the group file as space-separated instead of tab to get around the error when running the SAIGE-QTL RV pipeline, so trying this here.

Context: https://centrepopgen.slack.com/archives/C030X7WGFCL/p1725919664235749


Alexander-Stuckey (approved):
Looks good to me

PR #129, opened 2024-09-11
change ids to match with :

annacuomo:
Latest error suggests poor formatting, and previous errors (maybe) that IDs don't match between genotype file and group file, so addressing this here, reverting to `chr:pos:ref:alt` format


Alexander-Stuckey (approved):
Looks good to me

PR #134, opened 2024-09-13
only take var pos for distances

annacuomo:
Last run [failed](https://batch.hail.populationgenomics.org.au/batches/483632/jobs/1) because we changed the variant ids to not be just numerical.
Extracting numerical part of variant, i.e. the position from `chr:pos:ref:alt`

Syntax tested in a notebook:

Screenshot 2024-09-13 at 2 31 17 PM

Then I also noticed the genome reference wasn't used as an arg where it should be so, fixing that here too.

### DIFF
--- a/make_group_file.py
+++ b/make_group_file.py
@@ -101,7 +101,12 @@ def main(
             gene_interval = f'{num_chrom}:{window_start}-{window_end}'
             # extract variants within interval
             ds_result = hl.filter_intervals(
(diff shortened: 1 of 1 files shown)


ckrishk (approved):
looks good to me

annacuomo:
> looks good to me

thank you!

annacuomo:
running: https://batch.hail.populationgenomics.org.au/batches/483641/jobs/1

PR #137, opened 2024-09-18
feat: pas-446 create a vds subsetting script

Alexander-Stuckey:
Create a script that will take in a VDS path, an output format (although only VDS works right now) and optionally a number of samples and / or list of regions to subset down to.

The script will then subset down the VDS to those samples / regions, and output the requested output files + a sample list (if subset) to GCS.

### DIFF
--- a/subset_vds.py
+++ b/subset_vds.py
@@ -1,0 +1,373 @@
+"""Create a subset VDS from a larger VDS
+
+This script subsets a given VDS by either a number of samples, a region or set of regions, or both.
+
+Arguments:
+    --vds-path: str. Path to the VDS in GCS. Does not need the project. e.g cpg-bioheart-test/vds/bioheart1-0.vds should be entered as vds/bioheart1-0.vds
+    --n-samples: int, optional. The number of samples to subset the VDS down to.
+    --intervals: str, optional. A (comma separated) string in the format 'chr:start-end' of the interval to subset to, or the path to a file in gcs with one interval per line.
+    --output-format: str. A comma separated string of output formats to generate. Only formats in [vcf, bed, vds] can be chosen.
+
+Raises:
+    ValueError: only supports values of n-samples that are less than the total number of sampels in the input VDS.
+    FileExistsError: Will not overwrite existing files.
+    AttributeError: Unrecognised arguments.
+    AttributeError: The user must provide one of n-samples or intervals, optherwise the script will not do any subsetting.
+"""
+
+from argparse import ArgumentParser
+from io import StringIO
+from random import sample
+from re import fullmatch, split
+from typing import Any
+
+from cpg_utils.config import dataset_path, get_access_level, output_path, to_path
+from cpg_utils.hail_batch import init_batch
+from hail import (
+    IntervalExpression,
+    MatrixTable,
+    Table,
+    export_plink,
+    export_vcf,
+    get_reference,
+    parse_locus_interval,
+    split_multi_hts,
+)
+from hail.vds import filter_intervals, filter_samples, read_vds, to_dense_mt
+from hail.vds.variant_dataset import VariantDataset
+
+
+def check_output_already_exists(output_format: list[str], infile_name: str) -> None:
+    """Check for existence of output files
+
+    Args:
+        output_format (list[str]): list of output formats to generate files for
+        infile_name (str): base name of the dataset
+
+    Returns:
+        Tuple[bool, str]: A tuple to indicate if any output files already exist, and which ones they are.
+    """
+    output_errors: str = ""
+    files_exist_errors: bool = False
+    for format in output_format:
+        if format == "vds" and to_path(output_path(f"{infile_name}_subset")).exists():
+            output_errors += f"The output VDS {infile_name}_subset already exists. Refusing to overwrite it.\n"
+            files_exist_errors = True
+        if (
+            format == "bed"
+            and to_path(output_path(f"{infile_name}_subset.bed")).exists()
+        ):
+            output_errors += f"The output {to_path(output_path(infile_name))}_subset.bed fileset exists. Refusing to overwrite it.\n"
+            files_exist_errors = True
+        if (
+            format == "vcf"
+            and to_path(output_path(f"{infile_name}_subset.vcf.bgz")).exists()
+        ):
+            output_errors += f"The output file {to_path(output_path(infile_name))}_subset.vcf.bgz exists. Refusing to overwrite it.\n"
+            files_exist_errors = True
+    if files_exist_errors:
+        raise FileExistsError(f"{output_errors}")
+
+
+def parse_intervals(intervals: str) -> list[str]:
+    """read intervals from either the command line or a file
+
+    Args:
+        intervals (str): a string of either comma separated intervals, or a path to a file of intervals (one per line) in gcs
+
+    Raises:
+        ValueError: the intervals need to match the following pattern: ^[0-9cxymt]+ (does it start with some kind of contig identifier?)
+
+    Returns:
+        list[str]: a list of strings representing intervals
+    """
+    interval_list: list[str] = []
+    problem_intervals: list[str] = []
+    if to_path(intervals).exists():
+        with open(to_path(intervals), "rt") as interval_file:
+            interval_list = interval_file.readlines()
+    else:
+        interval_list = intervals.split(",")
+    for interval in interval_list:
+        if not fullmatch("^[0-9chrxymt]+", interval.split(":")[0].lower()):
+            problem_intervals.append(interval)
+    if len(problem_intervals) > 0:
+        raise ValueError(
+            f"The list of supplied intervals contains the following incorrectly formatted contigs: {[problem for problem in problem_intervals]}"
+        )
+    return interval_list
+
+
+def convert_intervals_to_locus(interval_list: list[str]) -> list[IntervalExpression]:
+    """convert a list of string intervals to interval expressions that hail understands for locus filtering
+
+    Args:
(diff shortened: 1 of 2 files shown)


annacuomo:
looks good! Would be cool to have the option to subset to a whole chromosome rather than an interval if possible? Or have a default interval?

Alexander-Stuckey:
> Also for SAIGE-QTL specifically, the index file associated with a VCF should be `.csi` rather than `.tbi` (tabix)!

Hail `export_vcf` doesn't support writing a `.csi` index 😞 . I figured that an index is better than no index, although it does mean that you'd have to generate your own.

subset_vds.py
```diff
@@ -0,0 +1,373 @@
...
+        if not fullmatch("^[0-9chrxymt]+", interval.split(":")[0].lower()):
+            problem_intervals.append(interval)
+    if len(problem_intervals) > 0:
+        raise ValueError(
+            f"The list of supplied intervals contains the following incorrectly formatted contigs: {[problem for problem in problem_intervals]}"
+        )
+    return interval_list
+
+
+def convert_intervals_to_locus(interval_list: list[str]) -> list[IntervalExpression]:
```
MattWellie, line 101: I think a lot of this code block/validation you would get for free with just calling `parse_locus_interval` - it tolerates `chr`, `chr:start-stop`, `chr1-chr2`, and would give you positional validation based on the reference.

If this was partially taken from [here](https://github.com/populationgenomics/analysis-runner/blob/main/scripts/subset_matrix_table.py#L150), then yes, I probably also wasted some time.

One piece of validation you might be missing here is checking for a `chr` prefix on the chromosome name, or a check that the chromosome name appears in the hail.reference.contigs

subset_vds.py
```diff
@@ -0,0 +1,373 @@
...
+            outdata.close()
+
+
+def main(
+    vds_path: str,
+    n_samples: int | None,
+    intervals: str | None,
+    output_formats: list[str],
+) -> None:
+    allowed_output_formats: list[str] = ["vcf", "bed", "vds"]
```
MattWellie, line 297: you can put this validation in the argparse interface
https://docs.python.org/3/library/argparse.html#choices

MattWellie: wait, no you can't... not if you tolerate multiple. Curses.

subset_vds.py
```diff
@@ -0,0 +1,373 @@
...
+        subset_vds.write(output_path(f"{infile_name}_subset", test=is_test))
+
+    subset_dense_mt: MatrixTable | Table | Any = to_dense_mt(subset_vds)
+    subset_dense_mt = split_multi_hts(subset_dense_mt)
+
+    for format in output_formats:
+        if format == "bed":
+            export_plink(
+                subset_dense_mt,
+                output_path(f"{infile_name}_subset", test=is_test),
```
MattWellie, line 263: `output_path` will handle this automatically - output_path builds an output location based on a few things in config:
1. default bucket for the dataset
2. output_prefix
3. file path/suffix

[output_path](https://github.com/populationgenomics/cpg-utils/blob/main/cpg_utils/config.py#L421) calls [dataset_path](https://github.com/populationgenomics/cpg-utils/blob/main/cpg_utils/config.py#L297), which builds the full path.

For a test run, the only paths in config are to test buckets, so it will only know how to build test output paths.

For main/full runs, there will be both test and main bucket paths in config. This `test` toggle is designed so that on a full/main permission run you can choose to direct the output to a test bucket anyway.

From the way you're using this (based on the access level), you'd be directing the output to the location it would be writing to by default, so the test argument doesn't really help you here.

subset_vds.py
```diff
@@ -0,0 +1,373 @@
...
+    n_total_samples: int = len(starting_samples)
+    if n_samples == 0:
+        raise ValueError(
+            "Subsetting to 0 samples will result in an empty MatrixTable, so not doing that."
+        )
+    if n_samples >= n_total_samples:
+        raise ValueError(
+            f"The subset sample size {n_samples} is at least the same size as the total number of samples in the input {n_total_samples}. Please choose a smaller number."
+        )
+    subset_sample_list = sample(starting_samples, n_samples)
```
MattWellie, line 185: do we want to seed this for reproducibility?

MattWellie (changes requested):
I think the `test=` specifier on output_path is redundant, but otherwise looks pretty hot

subset_vds.py
MattWellie: ooooh, or can you. You could have a list of choices, set `nargs='+'` on the argument, and take multiple individual arguments. that would remove the need for 
```output_formats=args.output_formats.lower().split(",")```

Comment (Alexander-Stuckey) [reply subset_vds.py:297]: Ah, that was what I was wanting to do, I assume then it's just multiple space separated arguments to --output-formats?

File: subset_vds.py (hunk 3 above)
Comment (Alexander-Stuckey) [reply subset_vds.py:263]: Makes sense, I can remove this

File: subset_vds.py (hunk 4 above)
Comment (Alexander-Stuckey) [reply subset_vds.py:185]: We can get this for free, so why not

File: subset_vds.py (hunk 1 above)
Comment (Alexander-Stuckey) [reply subset_vds.py:101]: I did use that as a reference, yes. Not familiar with using hail locuses yet.
If I can delete all this checking code and basically rely on `parse_locus_interval` to do all the checking for me, then I'm happy to do that. Thoughts?

Comment (MattWellie) [reply subset_vds.py:102]: It looks like you've already moved in that direction. Sounds good to me - if you just force a user to provide one of the Hail supported input formats, and if there are any invalid ones print a link to the Hail docs on which ones are valid, then it's still pretty flexible, and takes the burden off you writing the validation

File: subset_vds.py [hunk 5]
```diff
@@ -0,0 +1,340 @@
...
+        "--random-seed",
+        help="Seed for the random number generator when subsetting by samples",
+        required=False,
+        default=19700101,
+        type=int,
+    )
+    args, failures = parser.parse_known_args()
+    if failures:
+        parser.print_help()
+        raise AttributeError(f"Invalid arguments {failures}")
```
MattWellie, line 329: It's not important here, but I completely misunderstood this method, and was using it incorrectly.

`parser.parse_args()` attempts to parse all arguments, and will fail if there are additional arguments provided.

`parser.parse_known_args()` only parses the listed arguments, and will tolerate other arguments being present (it'll return the other un-digested ones). I'm not sure when you'd want to parse only a subset of the CLI arguments, but that would be permitted in parse_known_args. 

TL;DR - if you always want to fail when unknown arguments are present, the standard parse_args is the way to go

subset_vds.py
```diff
@@ -0,0 +1,340 @@
...
+        if to_path(output_path("subset_samples_file.txt")).exists():
+            raise FileExistsError(
+                f"The file {to_path(output_path('subset_samples_file.txt'))} exists. Refusing to overwrite it."
+            )
+        outdata = StringIO()
+        for single_sample in subset_sample_list:
+            outdata.write(f"{single_sample}\n")
+        with to_path(output_path("subset_samples_file.txt")).open("wt") as outfile:
+            outfile.write(outdata.getvalue())
+            outdata.close()
```
MattWellie, line 258:
```suggestion
        with to_path(output_path("subset_samples_file.txt")).open("wt") as outfile:
            outfile.write("\n".join(subset_sample_list))
```

You should just be able to write the string content directly without the StringIO intermediary

subset_vds.py
```diff
@@ -0,0 +1,340 @@
...
+            if "gvcf_info" in subset_dense_mt.entry:
+                subset_dense_mt = subset_dense_mt.drop("gvcf_info")
+            export_vcf(
+                subset_dense_mt,
+                output_path(f"{infile_name}_subset.vcf.bgz"),
+                tabix=True,
+            )
+
+    if subset_sample_list:
+        if to_path(output_path("subset_samples_file.txt")).exists():
```
MattWellie, line 249: This is a bit of a nit, but you might want to move this existence check earlier in the process, else you could potentially be in a position where the subset itself has been written, but you raise an exception before writing the sample list.

That's probably not going to happen, and you can pull the Sample subset easily from the VDS itself, so not a real issue

subset_vds.py
```diff
@@ -0,0 +1,340 @@
...
+        ValueError: if n_samples is >= to the number of samples in the input vds throw an error, as this would not do any filtering
+
+    Returns:
+        list[str]: a random list of samples of length n_samples from the input vds, to use in subsetting
+    """
+    starting_samples: list[str] = input_vds.variant_data.s.collect()
+    n_total_samples: int = len(starting_samples)
+    if n_samples == 0:
+        raise ValueError(
+            "Subsetting to 0 samples will result in an empty MatrixTable, so not doing that."
```
MattWellie, line 156: > so not doing that

yeah, you tell 'em

MattWellie (approved):

subset_vds.py
Alexander-Stuckey: I can, this just avoids multiple writes (one per line) to GCP, instead writing the whole file at once. I figured this'd be better practice

subset_vds.py
Alexander-Stuckey: Ah, I had assumed that the functionality was reversed, as I'd seen it used around CPG like that. I'll make that change

subset_vds.py
Alexander-Stuckey: Fair point though, changed that

subset_vds.py
MattWellie: Almost certainly me writing it/others copying my mistake 😆

MattWellie (approved):

PR #138, opened 2024-09-18
RV pipeline clean up

annacuomo:
Just removing commented out lines and updating running instructions to reflect recent changes (in `main` only because it can't run in `test` atm still)


Alexander-Stuckey (approved):
Looks good to me

PR #140, opened 2024-09-18
Update readme rv

annacuomo:
Ready for review!

PR updating the README to all the new changes, particularly around the SAIGE-QTL RV pipeline.

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,6 +204,15 @@ Set-based association testing ([rare variants step 2](https://weizhou0.github.io
 
 Instructions to run each component of the pipeline using analysis runner are provided at the top of each script.
 
+Briefly, if one wanted to run both common and rare variant pipelines, the order of running would be:
+
+1. get_genotype_vcf.py
+2. get_sample_covariates.py (does not require any other part of the pipeline and can be run in parallel with 1)
+3. get_anndata.py (requires 2)
+4. saige_assoc.py (requires 1,3)
+5. make_group_file.py (requires 1,3)
+6. saige_assoc_set_test.py (requires 1,3,4 (so that step1 is only run once) and 5)
+
 ## Data
 
 TenK10K is matched single-cell RNA-seq (scRNA-seq) and whole-genome sequencing (WGS) data from up to 10,000 individuals:
--- a/README.md
+++ b/README.md
@@ -93,6 +93,7 @@ Outputs
 
 Notes: option to include no weights or to compute weights that reflect the distance of each variant from the gene's transcription start site (`dTSS`).
 Using one of the flags below it is possible to additionally test using equal weights.
+We use no annotations for now (set to `null`).
 
 
 ## SAIGE-QTL association pipeline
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -54,74 +77,158 @@ Outputs:
...
-* ```variance_ratio_path```: path to variance ratio txt file generated in step 1
-* ```min_maf```: minimum minor allele frequency (MAF) (default: 0)
-* ```min_mac```: minimum minor allele count (MAC) (default: 5)
-* ```loco_bool```: boolean specifying whether leave-one-chromosome-out should be used (default: ```FALSE```)
-* ```n_markers```: internal parameter to batchify variants tested (default: 10000),
-* ```spa_cutoff```: internal parameter to do with the saddlepoint approximation, does not make much of a difference for us (default: 10000),
+In script (`saige_assoc.py`):
+
+* `vcf_file`: path to VCF file containing genetic variants to test.
+* `vcf_file_index`: corresponding .csi index file (not .tbi). check, just prefix?
```
Alexander-Stuckey, line 175: Have you checked if it can be just prefix? (ditto for the other places that this is mentioned)

README.md
```diff
@@ -54,74 +77,158 @@ Outputs:
...
+* `saige_output_file`: path to output file (check name).
```
Alexander-Stuckey, line 176: Have you checked the name? Or is this a warning to other users that they need to check the name? (Ditto for other mentions of check name)

README.md
```diff
@@ -54,74 +77,158 @@ Outputs:
...
+* `chrom`: chromosome to test.
+* `cis_window_file`: path to file specifying cis window / region to test (generated in part 2 of the pipeline, get anndata script).
+* `gmmat_model_path`: path to estimated null model (.rda) generated in step 1.
+* `variance_ratio_path`: path to variance ratio txt file generated in step 1.
+
+In config (under `[saige.sv_test]` in `saige_assoc_test.toml`):
+
+* `vcfField`: DS for dosages, GT for genotypes (default: `GT`).
+* `minMAF`: minimum minor allele frequency (MAF) (default: 0).
+* `minMAC`: minimum minor allele count (MAC) (default: 5).
```
Alexander-Stuckey, line 186: What happens when both values are set? Does it take the largest?

Alexander-Stuckey (changes requested):
Left a few comments that would be good to get sorted out first, to clarify the readme

README.md
annacuomo: thanks, I had been meaning to check this and then forgot 🤦‍♀️

README.md
annacuomo: good question, I assume so? I'll check.

README.md
annacuomo: all done now

README.md
annacuomo: updated but not very happy with the name, will fix that separately

README.md
annacuomo: all done

Alexander-Stuckey (approved):
Happy with the changes, looks good to me

PR #118, opened 2024-09-03
add max iter pcg flag

annacuomo:
Adding one more flag to the config (increasing iterations to convergence), testing now (https://batch.hail.populationgenomics.org.au/batches/482647) with 5,000 to see this if this is why [this job failed](https://batch.hail.populationgenomics.org.au/batches/479141/jobs/1002)

Update: it worked! 🎉 

step 1: https://batch.hail.populationgenomics.org.au/batches/482647/jobs/252
step 2: https://batch.hail.populationgenomics.org.au/batches/482647/jobs/253
step 3: https://batch.hail.populationgenomics.org.au/batches/482647/jobs/254
summarise: https://batch.hail.populationgenomics.org.au/batches/482647/jobs/347

PR #141, opened 2024-09-19
update output paths for rv also

annacuomo:
equivalent to https://github.com/populationgenomics/saige-tenk10k/pull/119 (which [succeeded](https://batch.hail.populationgenomics.org.au/batches/483776) except for one gene that requires more iteration for convergence, see https://github.com/populationgenomics/saige-tenk10k/pull/118) but for the RV pipeline


Alexander-Stuckey (approved):
Looks good to me

PR #142, opened 2024-09-20
Better path name

annacuomo:
Improved output filenames to avoid getting confused with the `output_path()` function.
Updated in readme as well


Alexander-Stuckey (approved):
Looks good to me

PR #143, opened 2024-09-20
Update instructions at top of scripts

annacuomo:
Update running instructions at the top of each script to some that definitely work, as just tested (tested for the 971 genes on chr2, cell type: B naive):

- `get_genotype_vcf.py`: [batch 483831](https://batch.hail.populationgenomics.org.au/batches/483831) 🎉 
- `get_sample_covariates.py`: [batch 483833](https://batch.hail.populationgenomics.org.au/batches/483833) 🎉 
- `get_anndata.py`: [batch 483841](https://batch.hail.populationgenomics.org.au/batches/483841) 🎉 
- `saige_assoc.py`: [batch 483877](https://batch.hail.populationgenomics.org.au/batches/483877) (default max iter PCG: 500, storage: 25GB)
  - follow up for the 9 genes that failed to converge, increasing max iter PCG to 5,000 (see https://github.com/populationgenomics/saige-tenk10k/pull/118): [batch 483926](https://batch.hail.populationgenomics.org.au/batches/483926)) 
  - still only succeeded for 3 genes, trying again with max iter PCG 10,000 👀 : [batch 484401](https://batch.hail.populationgenomics.org.au/batches/484401)
  - all 6 still failed, trying one last time with max iter PCG 50,000: [batch 485276](https://batch.hail.populationgenomics.org.au/batches/485276) (fail again, drop for now)
  - running with those 6 genes skipped, just so the summarising step will run: [batch 485325](https://batch.hail.populationgenomics.org.au/batches/485325)  🎉
- `make_group_file.py`: [batch 483881](https://batch.hail.populationgenomics.org.au/batches/483881) 
  - cancelled and re-running post parallelisation (https://github.com/populationgenomics/saige-tenk10k/pull/146): [batch 484396](https://batch.hail.populationgenomics.org.au/batches/484396)
  - 3 jobs failed with a weird Hail error and 11 were cancelled as a consequence of those failures + max concurrency, so running again to resolve those: [batch 485260](https://batch.hail.populationgenomics.org.au/batches/485260) 🎉 
- `saige_assoc_set_test.py`: [batch 485300](https://batch.hail.populationgenomics.org.au/batches/485300) (storage: 30GB)
  - tries to run step 1 for the 6 failures above, still fails (solution for now add to [drop gene in the config](https://github.com/populationgenomics/saige-tenk10k/blob/main/saige_assoc_test.toml#L5))
  - cancelled because it was costing a lot to open the files? Running again after Matt's efficiency improvement (see https://github.com/populationgenomics/saige-tenk10k/pull/147): [batch 485308](https://batch.hail.populationgenomics.org.au/batches/485308) 
  - final run removing those 6 failed genes, just so the summarising step will run: [batch 485323](https://batch.hail.populationgenomics.org.au/batches/485323)  🎉

Pipeline (both single variant and test set tests) ran from end-to-end, with final results at: `cpg-bioheart-main-analysis/saige-qtl/bioheart_n990_and_tob_n1055/output_files/240920/summary_stats` 😍

### DIFF
--- a/get_anndata.py
+++ b/get_anndata.py
@@ -4,31 +4,26 @@
 This script will
 
 - open anndata expression files
-- open covariate files
+- open cell and samples covariate files
 - create pheno_cov_files
 - create cis window files
 
 these files will be used as inputs for the
-SAIGE-QTL association pipeline.
+SAIGE-QTL association pipelines (both common and rare).
 
 To run:
 
-In main:
-
-In test:
-
 analysis-runner \
-    --description "make expression input files" \
-    --dataset "bioheart" \
-    --access-level "test" \
-    --output-dir "saige-qtl/bioheart_n990_and_tob_n1055/input_files" \
-    --image australia-southeast1-docker.pkg.dev/cpg-common/images/scanpy:1.9.3 \
-    python3 get_anndata.py --celltypes B_naive --chromosomes chr21 \
-    --anndata-files-prefix gs://cpg-bioheart-test/saige-qtl/anndata_objects_from_HPC \
-    --celltype-covs-files-prefix gs://cpg-bioheart-test/saige-qtl/celltype_covs_from_HPC \
-    --sample-covs-file gs://cpg-bioheart-test-analysis/saige-qtl/input_files/covariates/sex_age_geno_pcs_shuffled_ids_tob_bioheart.csv \
-    --concurrent-job-cap=1000 --pc-job-cpu=1 --cis-job-cpu=1
-
+   --description "make expression input files" \
+   --dataset "bioheart" \
+   --access-level "full" \
+   --output-dir "saige-qtl/bioheart_n990_and_tob_n1055/input_files/240920" \
+   --image australia-southeast1-docker.pkg.dev/cpg-common/images/scanpy:1.9.3 \
+   python3 get_anndata.py --celltypes B_naive --chromosomes chr2 \
+   --anndata-files-prefix gs://cpg-bioheart-main/saige-qtl/240-libraries/anndata_objects_from_HPC \
+   --celltype-covs-files-prefix gs://cpg-bioheart-main/saige-qtl/240-libraries/celltype_covs_from_HPC \
+   --sample-covs-file gs://cpg-bioheart-main-analysis/saige-qtl/input_files/240920/covariates/sex_age_geno_pcs_shuffled_ids_tob_bioheart.csv \
+   --pc-job-mem=8G
 
 """
 
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ Outputs:
 * VCF files containing all retained rare variants (one per chromosome) + corresponding index files (`.csi`)
 * plink object for only 2,000 variants (minor allele count > 20), after LD pruning - this is for the estimation of the variance ratio (VRE plink files)
 
-### Notes
+### Notes (get_genotype_vcf.py)
 
 SAIGE-QTL allows numeric chromosomes only, so both the .bim and the VCF files are modified in this script to remove the 'chr' notation (so that e.g., 'chr1' becomes '1').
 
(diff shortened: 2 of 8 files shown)


get_anndata.py
```diff
@@ -4,31 +4,26 @@
...
-    --image australia-southeast1-docker.pkg.dev/cpg-common/images/scanpy:1.9.3 \
-    python3 get_anndata.py --celltypes B_naive --chromosomes chr21 \
-    --anndata-files-prefix gs://cpg-bioheart-test/saige-qtl/anndata_objects_from_HPC \
-    --celltype-covs-files-prefix gs://cpg-bioheart-test/saige-qtl/celltype_covs_from_HPC \
-    --sample-covs-file gs://cpg-bioheart-test-analysis/saige-qtl/input_files/covariates/sex_age_geno_pcs_shuffled_ids_tob_bioheart.csv \
-    --concurrent-job-cap=1000 --pc-job-cpu=1 --cis-job-cpu=1
-
+   --description "make expression input files" \
+   --dataset "bioheart" \
+   --access-level "full" \
```
Alexander-Stuckey, line 19: Do you need `full` access, or is `main` enough? (Or whatever the level is called in between test and full)

annacuomo: That is one of the main questions I had! I used to always run everything with `standard`, get some permission error, try again with `full` and it would work! But I could never remember for which of the scripts that was, and as it turns out is every one of them (I think)?

It was definitely always true for the `get_sample_covariates.py` script (something to do with metamist) but now it's all scripts. Matt or others would know for sure but my guess of the culprit is [this type of statement ](https://github.com/populationgenomics/saige-tenk10k/blob/main/make_group_file.py#L163-L167) or in general lines checking for existence of the file before I run?

annacuomo: For the record, I'll list below the failed logs when running with `standard`:

* `get_genotype_vcf.py`: https://batch.hail.populationgenomics.org.au/batches/483824/jobs/1
* `get_sample_covariates.py`: https://batch.hail.populationgenomics.org.au/batches/485730/jobs/1
* `get_anndata.py`: https://batch.hail.populationgenomics.org.au/batches/485749/jobs/1

For the first two, it may be because I am wanting to access `test` objects, for the get anndata script it's because I'm wanting to copy to `tmp`?

@MattWellie would you be able to comment?

MattWellie: All pretty irritating issues - `test` only has access to `test`, `standard` has access to `main`, `full` has access to both. To resolve the need for `full` when running read-only processes that are looking at test data, you should probably move the data you're reading into `main` (this would also futureproof it by making it ~impossible to delete)

The `get_anndata.py` is really sucky - when you do a gsutil cp the client usually stages files (or file fragments locally) during the copy, then deletes the temp files. This isn't permitted with `standard` permissions, as that permission level can't delete any data from `main`, even from the `main-tmp` bucket. TBH this is a real user friction oversight, and we should probably resolve this by making `-tmp` buckets more like scratch space with lower/no write protection... In the short term, the solution is having to run with `full` permissions, or find an alternative form of the copy that doesn't attempt to clean up its own temp files :/

annacuomo: After discussing this morning with @MattWellie and @Alexander-Stuckey we agreed that for now there is no obvious solution for the `get_anndata.py` and SAIGE-QTL runner scripts (`saige_assoc.py` and `saige_assoc_set_test.py`), as they require copying and deleting from the `main-tmp` bucket.

I have updated the `get_genotype_vcf.py` script to take a file in `main-analysis` in with info on related individuals and have updated the instructions to require `standard` permissions only.

The `get_sample_covariates.py` script needs a little more work as I need to take in new `main-analysis` Hail Table files instead of temporary CSV / TSV (one of which in `test`) so I have done that in a separate PR (https://github.com/populationgenomics/saige-tenk10k/pull/151).

`make_group_file.py` ran fine in `standard`: https://batch.hail.populationgenomics.org.au/batches/485958 so updating that as well

Alexander-Stuckey (approved):
Looks good to me

Issue #71, opened 2024-03-04
get_genotype_vcf.py review


katiedelange:
Some points to consider in get_genotype_vcf.py

- [Extracting the chr values from the bim](https://github.com/populationgenomics/saige-tenk10k/blob/4f00b431e873f98f7c3890ebdb191fb80d975568/get_genotype_vcf.py#L136) (used to clean up the variants selected for variance ratio estimation) will leave empty values for the X, Y and MT chromosomes. However, you don't explicit exclude these chromosomes when creating the initial variant subset, and so you may end up with missing chr data being passed to SAIGE. I'd recommend either excluding these chromosomes from variant selection (if they are potentially going to skew variance ratio estimation), or adjusting your regex to `(\d+|X|Y|MT)`.

- I believe the current checkpointing approach (e.g. [here](https://github.com/populationgenomics/saige-tenk10k/blob/4f00b431e873f98f7c3890ebdb191fb80d975568/get_genotype_vcf.py#L258C1-L263C54)) is not going to save you anything (unless there's some hail magic I'm missing). Should you not check for existence of the checkpoint first and then run e.g. densification if and only if the checkpoint is not already available?

- You may wish to predicate the [downsampling to 1%](https://github.com/populationgenomics/saige-tenk10k/blob/4f00b431e873f98f7c3890ebdb191fb80d975568/get_genotype_vcf.py#L288) based on the number of variants left, to avoid accidentally downsampling to less than 2000 variants?

- Ideally [r2 and bp window](https://github.com/populationgenomics/saige-tenk10k/blob/4f00b431e873f98f7c3890ebdb191fb80d975568/get_genotype_vcf.py#L292) would be parameters you can adjust. Why did you choose these values in the first instance (NB they seem fine to me), and can you add that information to the README?

- Could you briefly explain the logic [behind this sampling approach](https://github.com/populationgenomics/saige-tenk10k/blob/4f00b431e873f98f7c3890ebdb191fb80d975568/get_genotype_vcf.py#L301-L302) / add it to the README?

- Just noting that you use MAF > Threshold. It's done consistently in the code and README, so is fine, but I feel like most studies I've seen include the lower threshold in their tested range (i.e. keep MAF >= threshold).

MattWellie:
> I believe the current checkpointing approach is not going to save you anything

[...]

PR #144, opened 2024-09-23
Address Katie's get_genotype_vcf.py issue (#71)

annacuomo:
Addressing @katiedelange's comments from https://github.com/populationgenomics/saige-tenk10k/issues/71

* LD pruning parameters as arguments
* pruning and downsampling approach described in the README
* exclude X,Y,MT chroms from VRE genotype file, and update README
* only subset (to 1%)  if necessary 
* reduce checkpoints (see discussion below), leaving only the one post subsetting and pre-pruning, since the subsetting may make the re-partitioning less than ideal, and the pruning is an expensive operation that doesn't do much internal checkpointing.

@katiedelange ready for you to review!
@MattWellie thanks for the comments below re: checkpoints 🙏

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,14 @@ Outputs:
 * VCF files containing all retained rare variants (one per chromosome) + corresponding index files (`.csi`)
 * plink object for only 2,000 variants (minor allele count > 20), after LD pruning - this is for the estimation of the variance ratio (VRE plink files)
 
-Notes: SAIGE-QTL allows numeric chromosomes only, so both the .bim and the VCF files are modified in this script to remove the 'chr' notation (so that e.g., 'chr1' becomes '1').
+### Notes
+
+SAIGE-QTL allows numeric chromosomes only, so both the .bim and the VCF files are modified in this script to remove the 'chr' notation (so that e.g., 'chr1' becomes '1').
+
+For the VRE estimation, we need to select 2,000 (this number can be customised) variables at random, except we need them to be common enough (MAC>20, also customisable), and we want them to be somewhat independent, to be more representative.
+To achieve this, we perform LD pruning of the variables so that variants in strong LD get pruned to one in each LD block.
+Because the LD pruning operation is very costly, we downsample the variants first (to 1% by default).
+The LD pruning parameters can be user-adjusted, with default values as described in the [methods's documentation](https://hail.is/docs/0.2/guides/genetics.html#pruning-variants-in-linkage-disequilibrium).
 
 ## Get sample covariates
 
--- a/README.md
+++ b/README.md
@@ -45,7 +45,8 @@ Outputs:
 
 SAIGE-QTL allows numeric chromosomes only, so both the .bim and the VCF files are modified in this script to remove the 'chr' notation (so that e.g., 'chr1' becomes '1').
 
-For the VRE estimation, we need to select 2,000 (this number can be customised) variables at random, except we need them to be common enough (MAC>20, also customisable), and we want them to be somewhat independent, to be more representative.
+For the VRE estimation, we need to select 2,000 (this number can be customised) variables at random, except we need them to be common enough (MAC>20, also customisable), and we want them to be somewhat independent, to be more representative of the entire cohort.
+We also subset to only variants on autosome chromosomes.
 To achieve this, we perform LD pruning of the variables so that variants in strong LD get pruned to one in each LD block.
 Because the LD pruning operation is very costly, we downsample the variants first (to 1% by default).
 The LD pruning parameters can be user-adjusted, with default values as described in the [methods's documentation](https://hail.is/docs/0.2/guides/genetics.html#pruning-variants-in-linkage-disequilibrium).
--- a/get_genotype_vcf.py
+++ b/get_genotype_vcf.py
@@ -218,6 +218,8 @@ def remove_chr_from_bim(input_bim: str, output_bim: str, bim_renamed: str):
     '--relateds-to-drop-path',
     default='gs://cpg-bioheart-test/large_cohort/v1-0/relateds_to_drop.ht',
 )
+@click.option('--ld-prune-r2', default=0.2)
+@click.option('--ld-prune-bp-window-size', default=500000)
 @click.command()
 def main(
     vds_path: str,
@@ -230,6 +232,8 @@ def main(
     exclude_indels: bool,
     plink_job_storage: str,
     relateds_to_drop_path: str,
+    ld_prune_r2: float,
+    ld_prune_bp_window_size: int,
 ):
     """
     Write genotypes as VCF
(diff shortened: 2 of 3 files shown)


MattWellie:
My gut feeling is that there are probably too many checkpoints in here at the moment. Each checkpoint writes the full MT/VDS out to disc, which is expensive & time consuming, as well as costing $ until it's deleted. Between the changes you're making you have 3 separate checkpoints, I'd try reducing that to just 1, probably just after the subsetting. From memory it was important to checkpoint just before running the LD-pruning method, as that doesn't do any checkpointing, but does add a lot of computational steps, so it's prone to running out of memory.

You might also want to consider your use of `checkpoint_mt` - I would advise forcing a new checkpoint each time. If you run this twice, and change the settings in between (e.g. relateds_to_drop_path), the run will detect the first checkpoint, read it in, and continue from there. By continuing from a previous checkpoint you're not exactly sure what you're continuing from... checkpoint_mt does a lot of resuming which might have been good for iterative development, but IMO it's much safer to always overwrite the previous checkpoint.

... an alternative is to include all the settings into the checkpoint name (maybe as a hash), so that when you resume from a checkpoint you resume from a previous run which was using the exact same settings, but that's probably a bit of a hassle

MattWellie:
Unrelated, it looks like this could create a bias towards the start of the genome:
https://github.com/populationgenomics/saige-tenk10k/blob/5931504572a76c12272f924ecc3bcae1f3ba8b2c/get_genotype_vcf.py#L388-L390

"subsample all variants based on a ratio of `n * 1.1`, then choose the first `n` rows"

I vaguely remember this being developed, but I don't remember the importance of the 1.1, or whether a slight regional bias would be important

annacuomo:
> Unrelated, it looks like this could create a bias towards the start of the genome:
> 
> https://github.com/populationgenomics/saige-tenk10k/blob/5931504572a76c12272f924ecc3bcae1f3ba8b2c/get_genotype_vcf.py#L388-L390
> 
> "subsample all variants based on a ratio of `n * 1.1`, then choose the first `n` rows"
> 
> I vaguely remember this being developed, but I don't remember the importance of the 1.1, or whether a slight regional bias would be important

So ideally here I'd just sample `n` exact variants but the `sample_rows` requires a portion instead (between 0 and 1). Because I wanted exactly `n` and that ratio may be just under I am multiplying but slightly more to make sure I get at least `n`, and then take the head. Any cleaner solutions would be more than welcome, but also a slight regional bias would not be problematic at all here imo!

annacuomo:
> My gut feeling is that there are probably too many checkpoints in here at the moment. Each checkpoint writes the full MT/VDS out to disc, which is expensive & time consuming, as well as costing $ until it's deleted. Between the changes you're making you have 3 separate checkpoints, I'd try reducing that to just 1, probably just after the subsetting. From memory it was important to checkpoint just before running the LD-pruning method, as that doesn't do any checkpointing, but does add a lot of computational steps, so it's prone to running out of memory.
> 
> You might also want to consider your use of `checkpoint_mt` - I would advise forcing a new checkpoint each time. If you run this twice, and change the settings in between (e.g. relateds_to_drop_path), the run will detect the first checkpoint, read it in, and continue from there. By continuing from a previous checkpoint you're not exactly sure what you're continuing from... checkpoint_mt does a lot of resuming which might have been good for iterative development, but IMO it's much safer to always overwrite the previous checkpoint.
> 
> ... an alternative is to include all the settings into the checkpoint name (maybe as a hash), so that when you resume from a checkpoint you resume from a previous run which was using the exact same settings, but that's probably a bit of a hassle

Hi @MattWellie thanks, I can't say I remember making that `checkpoint_mt` method, so apologies if I didn't appreciate how it was mostly for development! I have commented out most checkpoints now, only leaving the one before LD pruning, and commented out any use of `checkpoint_mt` as well. Could you check whether this is what you meant, and if so I'll go ahead and delete those lines instead! Thank you!

MattWellie:
> I can't say I remember making that `checkpoint_mt` method, and I still don't fully get it

I think I may have written it 😆 

It does 2 things - 
- creates a checkpoint
- writes a second checkpoint with repartitioned data

The repartitioning was inspired by problems I was running into in the RD pipeline - each separate shard of data is processed in parallel, if you have too many partitions, hail has to manage a lot of overhead organising all the jobs. If you have too few partitions, you don't get the benefits of parallel processing and the VM can run out of resources.

The issue it aims to solve is the content of each partition - if you start with well partitioned data (each partition has a roughly equal data content), then you apply a number of filters, this can unbalance your partitions - some being empty, some remaining full. As a specific example, when you down sample your data to 1%, the partition counts are unchanged, so you are making your partitions 99% more sparse. In this situation you probably want to run your down sampling, then repartition, so that instead of creating a huge number of VMs, each processing a trivial number of variants, you make a much smaller number of VMs, each doing a meaningful amount of computation.

Sorry if that's a bit of a ramble, but that's what this method was aiming to achieve. I think that's still reasonable based on this being a progressive filtering process, where repartitioning should give you meaningful improvements downstream (whether that's faster runtime, or just a process that succeeds instead of failing).

The bit I'm more concerned about is the detection of previous checkpoints, and resuming from them. If you only plan to run this once, or you only plan to run it with one set of settings, then don't worry about it.

If you're likely to run this more than once, you can run into an issue where you run once with a given set of settings/samples to keep/thresholds, then a subsequent run with different settings will continue from those checkpoints instead of writing new ones - if this script in its current form is run with different settings, it's hard to be confident about the data you're using.

Removing the read-existing-checkpoint behaviour, or always calling `checkpoint_mt` with `force=True` would solve this

annacuomo:
> > I can't say I remember making that `checkpoint_mt` method, and I still don't fully get it
> 
> I think I may have written it 😆
> 
> It does 2 things -
> 
> * creates a checkpoint
> * writes a second checkpoint with repartitioned data
> 
> The repartitioning was inspired by problems I was running into in the RD pipeline - each separate shard of data is processed in parallel, if you have too many partitions, hail has to manage a lot of overhead organising all the jobs. If you have too few partitions, you don't get the benefits of parallel processing and the VM can run out of resources.
> 
> The issue it aims to solve is the content of each partition - if you start with well partitioned data (each partition has a roughly equal data content), then you apply a number of filters, this can unbalance your partitions - some being empty, some remaining full. As a specific example, when you down sample your data to 1%, the partition counts are unchanged, so you are making your partitions 99% more sparse. In this situation you probably want to run your down sampling, then repartition, so that instead of creating a huge number of VMs, each processing a trivial number of variants, you make a much smaller number of VMs, each doing a meaningful amount of computation.
> 
> Sorry if that's a bit of a ramble, but that's what this method was aiming to achieve. I think that's still reasonable based on this being a progressive filtering process, where repartitioning should give you meaningful improvements downstream (whether that's faster runtime, or just a process that succeeds instead of failing).
> 
> The bit I'm more concerned about is the detection of previous checkpoints, and resuming from them. If you only plan to run this once, or you only plan to run it with one set of settings, then don't worry about it.
> 
> If you're likely to run this more than once, you can run into an issue where you run once with a given set of settings/samples to keep/thresholds, then a subsequent run with different settings will continue from those checkpoints instead of writing new ones - if this script in its current form is run with different settings, it's hard to be confident about the data you're using.
> 
> Removing the read-existing-checkpoint behaviour, or always calling `checkpoint_mt` with `force=True` would solve this

Great thanks for your help! If you have time can you review this PR from your perspective?

katiedelange (approved):
Looks good from my perspective, thanks for making those changes @annacuomo.

Issue #73, opened 2024-03-04
get_anndata.py review


katiedelange:
Some points to consider in get_anndata.py

- It's probably worth highlighting in the README that missing values for the covariates [will get filled in](https://github.com/populationgenomics/saige-tenk10k/blob/c6f09ee5d38023b30d59614422e118dd0a540117/get_anndata.py#L132-L135) as 0 ('unknown', correct?) for sex and the mean for age, rather than being left as missing. 

- It's probably a pretty minor computational cost, but just noting that you [fill in this missing data per sample](https://github.com/populationgenomics/saige-tenk10k/blob/c6f09ee5d38023b30d59614422e118dd0a540117/get_anndata.py#L123-L135) many many thousands of times, instead of doing it as a preprocessing step once and then reading in the cleaned up file. 

- Are missing values in the cell covariates file also cleaned up, i.e. in the upstream pipeline?

- Just a comment: there's a bit of an interesting point here about how lowly expressed ≠ missing (necessarily). While it makes sense to drop these genes and not bother testing them in the celltypes where they're lowly expressed, your final output is going to be non-square and it's worth remembering that this filter was applied when you interpret the results.

annacuomo:
@katiedelange 

> It's probably worth highlighting in the README that missing values for the covariates [will get filled in](https://github.com/populationgenomics/saige-tenk10k/blob/c6f09ee5d38023b30d59614422e118dd0a540117/get_anndata.py#L132-L135) as 0 ('unknown', correct?) for sex and the mean for age, rather than being left as missing.

Unless I got this wrong in the code, this is not the default behaviour but only if those flags for filling values are set to True? But yes I'll add a note in the readme about this!

annacuomo:
@katiedelange 

> It's probably a pretty minor computational cost, but just noting that you [fill in this missing data per sample](https://github.com/populationgenomics/saige-tenk10k/blob/c6f09ee5d38023b30d59614422e118dd0a540117/get_anndata.py#L123-L135) many many thousands of times, instead of doing it as a preprocessing step once (i.e. in [get_sample_covariates.py](https://github.com/populationgenomics/saige-tenk10k/blob/main/get_sample_covariates.py))and then reading in the cleaned up file.

Linked to the above, I think I was not overly concerned about this because my intention was 100% to not fill in these values by default, and I mostly added this because there was a lot of missingness in test which was causing some issues.
[...]

PR #145, opened 2024-09-23
Address Katie's get_anndata.py issue (#73)

annacuomo:
Addressing Katie's comments from https://github.com/populationgenomics/saige-tenk10k/issues/73

* update README to specify sex and age missing value choices
* update README to point out lowly-expressed genes are filtered out and won't even be tested
* option to fill in missing values for sex and age moved from the `get anndata` to the `get sample covariate` script (they were not provided as args before so it could not have been run before anyways 😅 )
  * testing get sample covs: https://batch.hail.populationgenomics.org.au/batches/485681 🎉 
  * testing get anndata: https://batch.hail.populationgenomics.org.au/batches/485696 🎉 

@katiedelange all done!

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ Outputs:
 
 * TSV sample covariate file (one per cohort)
 
-### Notes
+### Notes (get_genotype_vcf.py)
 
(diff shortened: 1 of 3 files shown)


katiedelange (approved):
Looks good, thanks for making those changes :)

PR #146, opened 2024-09-23
first attempt at parallelising make group file

annacuomo:
When [running](https://batch.hail.populationgenomics.org.au/batches/483881) for a larger chromosome (chrom 2, 971 genes), it is immediately apparent that this needs to be parallelised, rather than just running each gene in series 🤦‍♀️ 

attempting that in this PR, similar to what I do in the `get_anndata.py` script ([e.g.](https://github.com/populationgenomics/saige-tenk10k/blob/main/get_anndata.py#L348-L363)), 

testing in `test`: https://batch.hail.populationgenomics.org.au/batches/483974 🎉 

successful, now after adding max concurrency: https://batch.hail.populationgenomics.org.au/batches/483994 🎉 

(files in: `cpg-bioheart-test/saige-qtl/input_files/group_files/test_parallel/chr22`)

### DIFF
--- a/make_group_file.py
+++ b/make_group_file.py
@@ -45,7 +45,7 @@ def make_group_file(
     chrom: str,
     cis_window_files_path: str,
     # group_files_path: str,
-    group_file: str,
+    group_file,
     cis_window: int,
     genome_reference: str,
     gamma: str,
--- a/make_group_file.py
+++ b/make_group_file.py
@@ -29,9 +29,9 @@ analysis-runner \
 import click
 import logging
 
-import hailtop.batch.job as hb_job
+from typing import List
 
-import pandas as pd
+import hailtop.batch.job as hb_job
 
 from cpg_utils import to_path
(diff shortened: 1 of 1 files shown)


make_group_file.py
```diff
@@ -29,14 +29,85 @@
 import click
 import logging
 
-import math
+from typing import List
```
MattWellie, line 32: depending on how urgent this PR is, you can remove this. Type hinting hasn't required an imported `List` (Upper-case) for a few versions now. You can just use the lower-case `list` in type hints.

MattWellie (approved):
one comment

make_group_file.py
annacuomo: welcome back! Not super urgent, thanks just made the change!

annacuomo: testing just in case: https://batch.hail.populationgenomics.org.au/batches/484392

PR #147, opened 2024-09-24
Peak efficiency

MattWellie:
There's a chance this approach is much more resource efficient
There's also a chance it doesn't work at all

Changes: 

1. Instead of creating one 'part 2b' job per gene, we create once per Chromosome
2. we set a number `jobs_per_vm` representing the max number of jobs we try to run per image, and `jobs_in_vm` as a counter of the number of commands stacked inside the current VM
3. `build_run_set_based_test_command` now takes a job and a chromosome VCF as arguments, and adds an additional command
4. the output per gene/celltype is written to a unique local path, then moved to a unique resource group path (this... might not be necessary. I expect writing to `job.output` multiple times in a loop would go wrong, so I'm making a unique path each time)
5. After a new job has been added, we check if there are too many jobs in that VM - if so, create a new job

### DIFF
--- a/saige_assoc_set_test.py
+++ b/saige_assoc_set_test.py
@@ -149,7 +149,7 @@ def build_run_set_based_test_command(
     # write the output
     get_batch().write_output(second_job.output, set_output_path)
 
-    return second_job, second_job.output
+    return second_job
 
 
 def apply_job_settings(job: hb.batch.job.Job, job_name: str):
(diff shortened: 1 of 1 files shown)


saige_assoc_set_test.py
```diff
@@ -364,23 +388,37 @@ def manage_concurrency_for_job(job: hb.batch.job.Job):
...
+                step2_job.depends_on(gene_dependency)
 
                 # add this job to the list of jobs for this cell type
-                if gene_dependency is not None:
-                    celltype_jobs.setdefault(celltype, []).append(gene_dependency)
+                celltype_jobs.setdefault(celltype, []).append(step2_job)
+
+                # check if we need a new VM, i.e. we've hit the jobs-per-VM limit
+                if jobs_in_vm >= jobs_per_vm:
+                    step2_job = create_a_2b_job()
```
Alexander-Stuckey, line 421: Just feel like we should create a new 2b job here and then set `jobs_in_vm` back to `0`? Hope I'm following the code correctly.

MattWellie: yup, I forgot to reset the job counter

PR #148, opened 2024-09-24
Write to a valid local path

MattWellie:
Please have a looky looky, see if it would resolve the current errors


Alexander-Stuckey (approved):
Looks good to me

PR #149, opened 2024-09-24
More efficiencies

MattWellie:
Having a go at making the same changes to saige_assoc.py

1. Reads in the input VCF once per chromosome
2. One VM runs the SingleVariantTest for multiple separate genes
3. If the single VM exceeds the threshold number of jobs, start a new one

Also simplifies some of the syntax changes in the other script

I think this is pretty equivalent to #147, but it's for a job in the middle of the flow so may have slight differences

### DIFF
--- a/saige_assoc.py
+++ b/saige_assoc.py
@@ -371,9 +371,6 @@ def main(
 
     for chromosome in chromosomes:
 
-        single_var_test_job = create_second_job()
-        jobs_in_vm = 0
-
         # genotype vcf files are one per chromosome
         if test_str:
             vcf_file_path = (
@@ -383,6 +380,9 @@ def main(
             vcf_file_path = (
                 f'{genotype_files_prefix}/{chromosome}_common_variants.vcf.bgz'
             )
+
+        single_var_test_job = create_second_job(vcf_file_path)
+        jobs_in_vm = 0
 
         # read in vcf file once per chromosome
         vcf_group = get_batch().read_input_group(vcf=vcf_file_path, index=f'{vcf_file_path}.csi')
@@ -476,7 +476,7 @@ def main(
                     celltype_jobs.setdefault(celltype, []).append(job3)
 
                 if jobs_in_vm >= jobs_per_vm:
-                    single_var_test_job.depends_on(jobs[-jobs_per_vm])
+                    single_var_test_job = create_second_job(vcf_file_path)
                     jobs_in_vm = 0
 
     # summarise results (per cell type)
--- a/saige_assoc.py
+++ b/saige_assoc.py
@@ -56,6 +56,7 @@ import click
 
 import logging
 
+from google.cloud import storage
 import hailtop.batch as hb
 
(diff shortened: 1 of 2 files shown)


annacuomo (approved):

MattWellie:
@annacuomo I spotted an error in this, so I've patched that

I've also included a reactive VM sizing - size of the GCP VCF object + 5GB buffer for all the gene data

Alexander-Stuckey (approved):
Looks good to me

annacuomo:
Looks good to me, I'll merge and test now 🤞

annacuomo:
@MattWellie the writing two outputs almost worked but not quite:

the new output single assoc are now written appropriately, e.g.: `cpg-bioheart-main-analysis/saige-qtl/bioheart_n990_and_tob_n1055/output_files/240920/B_naive/chr1/B_naive_ENSG00000000457_cis_set.singleAssoc.txt`

the previous one though gets an extra `.set` at the end? E.g.,
`cpg-bioheart-main-analysis/saige-qtl/bioheart_n990_and_tob_n1055/output_files/240920/B_naive/chr1/B_naive_ENSG00000000457_cis_set.set`

MattWellie:
I thought that might happen.. is that something you can tolerate? In a declared resource group all the keys are the expected extensions, so when the group is all written out, they're used as extensions at the end of the file. These can't be empty...

If you need this to have no extension you can run write_output on each individual member of the resource group separately, i.e. writing the file `set` to the path without extension, and writing the `singleAssoc.txt` file to path + '.singleAssoc.txt'

annacuomo:
> I thought that might happen.. is that something you can tolerate? In a declared resource group all the keys are the expected extensions, so when the group is all written out, they're used as extensions at the end of the file. These can't be empty...
> 
> If you need this to have no extension you can run write_output on each individual member of the resource group separately, i.e. writing the file `set` to the path without extension, and writing the `singleAssoc.txt` file to path + '.singleAssoc.txt'

Ok! The way I see it these are our options:

1. minimal effort, change [this line](https://github.com/populationgenomics/saige-tenk10k/blob/main/saige_assoc_set_test.py#L248) to summarise files with `...cis_set.set` fairly ugly with the repetition of `set` but can probably live with it
2. better would be to update the term `set` to `rv`, e.g. [here](https://github.com/populationgenomics/saige-tenk10k/blob/main/saige_assoc_set_test.py#L391), technically the `singleAssoc.txt` results are not from a set test anyways, so then these would be rare variant (rv) set test, and rv single variant test. I'd maybe want to rename `set_key` to `rv_key` just to be consistent but I don't see any other implication?
3. do as you suggest writing the outputs separately

thoughts? If doing 2 for consistency we could update the saige_assoc.py step 2 output from `_cis` to `_cis.cv` but maybe not necessary

annacuomo:
@MattWellie attempt at option 2: https://github.com/populationgenomics/saige-tenk10k/pull/156

PR #150, opened 2024-09-25
add cis gene job mem also

annacuomo:
Recent cis gene file making jobs failed with an OOM error (e.g. https://batch.hail.populationgenomics.org.au/batches/485954/jobs/91), so adding option to increase mem here.


Alexander-Stuckey (approved):
Looks good to me

PR #151, opened 2024-09-25
New sex files

annacuomo:
Instead of using temporary files we previously created ad hoc to obtain sex information, use most up-to-date info from the Sample QC part from the large cohort pipeline (note: these are Hail Table files, rather than CSV/TSV).

At the moment, the more reliably up-to-date files are cohort specific (@katiedelange is this right?), but eventually I anticipate a single `tenk10k` file so I have added that syntax (commented out) for the future.

Using these files which are in `main-analysis` also allows us to run with `standard` permission, see discussion in this PR: https://github.com/populationgenomics/saige-tenk10k/pull/143

### DIFF
--- a/get_sample_covariates.py
+++ b/get_sample_covariates.py
@@ -13,36 +13,23 @@ these will be used in the gene expression input preparation
 building inputs for the SAIGE-QTL association pipeline.
 
 
-To run,
-
-in test:
+To run:
 
 analysis-runner \
-    --description "get sample covariates" \
-    --dataset "bioheart" \
-    --access-level "test" \
-    --output-dir "saige-qtl/input_files/covariates/" \
-    python3 get_sample_covariates.py --tob-sex-file-path 'gs://cpg-bioheart-test/saige-qtl/input_files/mapping_for_anna.csv' \
-                --bioheart-sex-file-path 'gs://cpg-bioheart-test-analysis/hoptan-str/somalier/somalier.samples.tsv' \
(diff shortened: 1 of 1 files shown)


get_sample_covariates.py
```diff
@@ -98,29 +85,37 @@ def main(
...
-    tob_meta['sample_id'] = tob_meta['new_CPG_id']
-    tob_sex = tob_meta.loc[:, ["sample_id", "sex"]]
-
-    # extract sex for BioHEART
-    bioheart_sex = bioheart_meta.loc[:, ["sample_id", "sex"]]
+    # option 1: separate files
+    # tob
+    tob_sample_qc_ht = hl.read_table(tob_sampleqc_file_path)
+    # convert to pandas
+    tob_sample_qc_df = tob_sample_qc_ht.to_pandas()
```
Alexander-Stuckey, line 92: Why read the table with hail and then convert to pandas, rather than just read with pandas to start? @MattWellie anything I'm missing here?

annacuomo: ooo did not know you could do that?!

MattWellie: From memory the input file is in a proprietary Hail format, so a read + convert with Hail was required, it's not a simple T/CSV

annacuomo: Naive try in a notebook didn't succeed, but happy to try things?

Screenshot 2024-09-25 at 11 57 59 AM

Alexander-Stuckey (approved):
Looks good to me

PR #152, opened 2024-09-25
add single variant results rare test

annacuomo:
So SAIGE-QTL rare variant test actually allows for running single-variant tests as well as set-based tests, when setting the `is_single_in_groupTest` flag to True ([which we are doing](https://github.com/populationgenomics/saige-tenk10k/blob/main/saige_assoc_test.toml#L51)).
The file gets saved to the same location where the set-based test results are written to, and has the same name except with an additional `.singleAssoc.txt` at the end (see [docs](https://weizhou0.github.io/SAIGE-QTL-doc/docs/set_step2.html)).

I've tried to modify this script to mimic something similar we do for the null_job output ([lines](https://github.com/populationgenomics/saige-tenk10k/blob/main/saige_assoc_set_test.py#L211-L216)) but I am really not confident 😅

### DIFF
--- a/saige_assoc_set_test.py
+++ b/saige_assoc_set_test.py
@@ -392,6 +392,13 @@ def main(
                 # unique output path for this set-based test
                 set_output_path = output_path(set_key, 'analysis')
 
+                step2_job.declare_resource_group(
+                    output={
+                        'set': '{root}',
+                        'singleAssoc.txt': '{root}.singleAssoc.txt',
+                    }
+                )
+
                 # if the output exists, do nothing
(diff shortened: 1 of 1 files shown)


MattWellie:
@annacuomo  I think this is the change you want to make: https://github.com/populationgenomics/saige-tenk10k/pull/153

Instead of catching one output one file, output two, and write them to the same location

annacuomo:
Closing this now as it should have been addressed in https://github.com/populationgenomics/saige-tenk10k/pull/153 then merged into https://github.com/populationgenomics/saige-tenk10k/pull/149

PR #154, opened 2024-09-26
output data to analysis

Alexander-Stuckey:
Output the subsets to the analysis bucket instead of the main bucket


MattWellie (approved):
LGTMatt

PR #156, opened 2024-09-26
Rename filepath from set to rare

annacuomo:
Implementing option 2 from here: https://github.com/populationgenomics/saige-tenk10k/pull/149#issuecomment-2376035205


MattWellie (approved):
Looks good!

PR #157, opened 2024-09-26
add common or rare flag

annacuomo:
The rare variant pipeline allows to run a single variant test also, and the results are identical to the outputs of the common variant pipeline, just written to a file that's named differently.

Here, I:

- add a flag that looks at the appropriate filename format depending on whether I am looking at common or rare results
- allow to run for multiple cell types
- remove the top SNP plots which aren't super informative
- update to current format of outputs


MattWellie (approved):
seems reasonable

PR #158, opened 2024-09-27
start adding to writeout file writing

annacuomo:
The idea here is that every time I run this pipeline I write out a text file detailing all the flags / files / params I ran this with, to aid reproducibility.

Specifically, included in the file are:
- paths to the input files (genotype files, phenotype + covariate files, cis window files)
- flags + params from config file (e.g. chromosomes and cell types considered, and then all [SAIGE-QTL](https://github.com/weizhou0/qtl) flags)
- `ar_guid`: analysis runner unique ID
- `runtime_config` is the combination of all the default analysis-runner content (references, images, etc.), plus any config files added using `--config`
- the filename will include date and time on when this is run, and the prefix can be provided as an argument to the script, so may include additional info on the "type" of run, e.g. "run with permuted ID to assess calibration", "test run on only one chrom", and maybe in the future "conditional analysis" or other!

### DIFF
--- a/saige_assoc.py
+++ b/saige_assoc.py
@@ -366,6 +366,10 @@ def main(
     vre_plink_path = f'{vre_files_prefix}/vre_plink_2000_variants'
     writeout_dict['vre_plink_files_prefix_used'] = vre_plink_path
 
+    writeout_dict['cis_window_files_path_used'] = cis_window_files_path
+    cis_window_size = get_config()['saige']['cis_window_size']
+    writeout_dict['cis_window_size_used'] = cis_window_size
+
     for chromosome in chromosomes:
 
         # genotype vcf files are one per chromosome
@@ -388,8 +392,6 @@ def main(
 
         # cis window files are split by gene but organised by chromosome also
         cis_window_files_path_chrom = f'{cis_window_files_path}/{chromosome}'
-
-        cis_window_size = get_config()['saige']['cis_window_size']
 
         for celltype in celltypes:
             # extract gene list based on genes for which we have pheno cov files
--- a/saige_assoc.py
+++ b/saige_assoc.py
@@ -359,18 +359,16 @@ def main(
 
     vre_plink_path = f'{vre_files_prefix}/vre_plink_2000_variants'
 
-    cis_window_size = get_config()['saige']['cis_window_size']
-
     # populate all the important params into a file for long-term reference
     writeout_dict: dict = {
         'ar_guid': try_get_ar_guid() or 'UNKNOWN',
-        'chromosomes': chromosomes,
-        'celltypes': celltypes,
-        'drop_genes': drop_genes,
         'vre_plink_files_prefix_used': vre_plink_path,
         'pheno_cov_files_path_used': pheno_cov_files_path,
         'cis_window_files_path_used': cis_window_files_path,
-        'cis_window_size_used': cis_window_size,
+        'saige_fit_null_params': get_config()['saige'],
+        'saige_single_variant_test_params': get_config()['saige.sv_test'],
+        'saige_fit_null_job_specs': get_config()['saige.job_specs.fit_null'],
+        'saige_single_variant_test_job_specs': get_config()['saige.job_specs.sv_test'],
         'runtime_config': getenv('CPG_CONFIG_PATH'),
     }
 
--- a/saige_assoc.py
+++ b/saige_assoc.py
@@ -316,7 +316,9 @@ def create_second_job(vcf_path: str) -> hb.batch.job.Job:
 @click.option(
     '--vre-files-prefix', default=dataset_path('saige-qtl/input_files/genotypes')
 )
-@click.option('--writeout-file-prefix', default=dataset_path('saige-qtl'))
+@click.option(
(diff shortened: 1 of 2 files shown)


annacuomo:
@MattWellie happy to save to `.json` if you think it's better since this is really only for reference, but just noting that I find json files less nice to read / less user friendly in case it was still possible to save to CSV or other?

annacuomo:
@MattWellie ah also one last thing I want to save is the output path where the results will be written to which I think is just `output_path`? is that part of analysis runner? How would you do this?

MattWellie:
> in case it was still possible to save to CSV or other?

You can save it to whatever you'd like! For this output your JSON file would look like:

```
{
  'key': 'value',
  'key2': ['value', 'value', 'value'],
  ...
```

Which is the most intuitive format possible, especially as some of your parameters are lists of strings - parsing comma separated values from a CSV is a mess 😬 

TOML is another format where you can store the entire dictionary with all the objects in their native formats (strings, numbers, lists, nested dictionaries...). If CSV/TSV is what you'd prefer, no dramas.

>  I want to save is the output path where the results will be written to which I think is just output_path

To get the generic output location you can just use `output_path('')` (i.e. the output_path location without an additional file path.

annacuomo:
> To get the generic output location you can just use output_path('') (i.e. the output_path location without an additional file path.

Ok should've done with the latest [commit](https://github.com/populationgenomics/saige-tenk10k/pull/158/commits/96da7d2658ae5e21c0c6b79fb844a67f31afd238)? Is this what you meant?

annacuomo:
> TOML is another format where you can store the entire dictionary with all the objects in their native formats (strings, numbers, lists, nested dictionaries...). If CSV/TSV is what you'd prefer, no dramas.

I guess as long as it's not super nested, the JSON file is still fairly readable? I don't know I am just not used to it I think, always find it hard to parse with all the `{`s 🤷‍♀️ 

I do quite like the TOML as in the config file? But honestly if JSON is what's best here all good

saige_assoc.py
```diff
@@ -341,12 +347,34 @@ def manage_concurrency_for_job(job: hb.batch.job.Job):
             job.depends_on(jobs[-get_config()['saige']['max_parallel_jobs']])
         jobs.append(job)
 
+    # define writeout file by type of pipeline and date and time
+    date_and_time = datetime.today().strftime('%Y-%m-%d_%H:%M:%S')
+    writeout_file = (
+        f'{writeout_file_prefix}/saige_qtl_common_variant_pipeline_{date_and_time}.csv'
```
MattWellie, line 353:
```suggestion
        f'{writeout_file_prefix}/saige_qtl_common_variant_pipeline_{date_and_time}.json'
```

sorry, this one is on me. The JSON library won't have any issues writing to a `.csv` file, but it will be misleading.

saige_assoc.py
```diff
@@ -313,6 +316,7 @@ def create_second_job(vcf_path: str) -> hb.batch.job.Job:
 @click.option(
     '--vre-files-prefix', default=dataset_path('saige-qtl/input_files/genotypes')
 )
+@click.option('--writeout-file-prefix', default=dataset_path('saige-qtl'))
```
MattWellie, line 319: I just had a thought... you probably want the default location for log files to be in `-analysis`, so that you won't have any issues reading them in future

saige_assoc.py
```diff
@@ -341,12 +347,34 @@ def manage_concurrency_for_job(job: hb.batch.job.Job):
...
+        'ar_guid': try_get_ar_guid() or 'UNKNOWN',
+        'results_output_path': output_path(''),
+        'vre_plink_files_prefix_used': vre_plink_path,
+        'pheno_cov_files_path_used': pheno_cov_files_path,
+        'cis_window_files_path_used': cis_window_files_path,
+        'saige_fit_null_params': get_config()['saige'],
+        'saige_single_variant_test_params': get_config()['saige.sv_test'],
+        'saige_fit_null_job_specs': get_config()['saige.job_specs.fit_null'],
+        'saige_single_variant_test_job_specs': get_config()['saige.job_specs.sv_test'],
+        'runtime_config': getenv('CPG_CONFIG_PATH'),
```
MattWellie, line 376: FYI the runtime config is the combination of all the default analysis-runner content (references, images, etc.), plus any config files you added using `--config` - it's the file that you read when you use `get_config`

MattWellie (changes requested):
one path name change, and I'd recommend shifting this to -analysis

Both things I should have spotted before 😅

saige_assoc.py
annacuomo: ah wait does this mean all of the the flags im adding using get_config() will be added twice?

MattWellie: Nope! This is just the path to that file, not the contents! In your output file you'll have some select entries, and the path to this file (if for some reason you wanted to check all the contents in that file)

MattWellie (approved):

saige_assoc.py
annacuomo: ok perfect! And it will be the actual file in the way it was saved when run, not what's committed in the repo at that stage right? And sorry the AR guid thing exactly what does it tell me? Will it know what commit / branch something was run with?

PR #159, opened 2024-09-27
start adding to conditional analysis

annacuomo:
Eventually this would implement a [conditional analysis](https://weizhou0.github.io/SAIGE-QTL-doc/docs/single_step2.html#conditional-analysis) (still super early stages)


annacuomo:
closing this as it looks super messy with all of the new changes from other PRs making this look more complicated than it needs to be.

Starting a fresh one at https://github.com/populationgenomics/saige-tenk10k/pull/166

PR #160, opened 2024-09-30
tweak tweak

MattWellie:
- include the AR GUID to help find the driver & associated other jobs for the batch
- include the path to the config file we draw from using get_config
- minimise the separate additions to the dictionary
- write using to_path/JSON, so we don't need to alter any formats (e.g. storing a list of strings, instead of needing to compress into a single string)

### DIFF
--- a/saige_assoc.py
+++ b/saige_assoc.py
@@ -32,17 +32,17 @@ analysis-runner \
 """
 
 import click
-
+import json
 import logging
-import pandas as pd
 
 from datetime import datetime
+from os import getenv
 
 from google.cloud import storage
 import hailtop.batch as hb
 
 from cpg_utils import to_path
-from cpg_utils.config import get_config, image_path, output_path
+from cpg_utils.config import get_config, image_path, output_path, try_get_ar_guid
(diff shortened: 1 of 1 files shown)


annacuomo:
Thanks! Is it possible to just extract all config from e.g. `get_config()['saige']` rather than spelling each of them out separately?

MattWellie:
> Thanks! Is it possible to just extract all config from e.g. `get_config()['saige']` rather than spelling each of them out separately?

100% If you're writing out a JSON you could even just put the whole of get_config in there, they're only dictionaries.

To keep it specific if you just want to focus on the saige params you could do 
```
     'saige_params': get_config()['saige'],
```

annacuomo:
> > Thanks! Is it possible to just extract all config from e.g. `get_config()['saige']` rather than spelling each of them out separately?
> 
> 100% If you're writing out a JSON you could even just put the whole of get_config in there, they're only dictionaries.
> 
> To keep it specific if you just want to focus on the saige params you could do
> 
> ```
>      'saige_params': get_config()['saige'],
> ```

ok will try and ask for PR again

PR #161, opened 2024-09-30
Readme and toml updates

annacuomo:
Fixing / updating the README and config files 

README:
* I had gotten the equal sign in the thresholds for common and rare variant definitions wrong 
* general prettifying / typos fixes / updates
* added abbreviations

TOML file for SAIGE-QTL running:
* default MAC for ultra-rare definition changed from 10 to 1 (makes more sense for this smaller cohort)
* increasing the max convergence to 5,000 (it won't matter for the vast majority of genes that won't have this issue, and it will rescue a few -- see https://github.com/populationgenomics/saige-tenk10k/pull/118)
* `cohort` is a categorical variable that was turned into dummy variables `tob` and `bioheart`, adding one here as the other will be perfectly collinear with it


katiedelange (approved):
Looks good to me.

PR #164, opened 2024-10-01
Update demux script

annacuomo:
Update this script to take paths and common variant thresholds as inputs, move to helper.

For context, this is somewhat separate from the main pipeline, it is a script that generates a common SNP VCF file that gets transferred over to HPC so we can demultiplex (assign cells to their donor of origin in a pool experiment, using [vireo](https://genomebiology.biomedcentral.com/articles/10.1186/s13059-019-1865-2)) our scRNA-seq data.

This happens upstream of the pipeline, so it goes:
* scRNA-seq processing up to demultiplexing (over at HPC)
* in parallel, WGS processing (large cohort pipeline)
* this script
* scRNA-seq can resume processing, starting with demultiplexing, up to making the [input files](https://github.com/populationgenomics/saige-tenk10k/blob/main/get_anndata.py#L190-L194) required for the SAIGE-QTL pipeline
* SAIGE-QTL pipeline can run (this repo)

Long slack thread for when this was first run for additional context: https://centrepopgen.slack.com/archives/C018KFBCR1C/p1703025906749599


MattWellie (approved):
Given that this is a rename and a change the GitHub diff is... not useful. Happy to rubber stamp this

annacuomo:
> Given that this is a rename and a change the GitHub diff is... not useful. Happy to rubber stamp this

I know right?! Should I have done something different? I don't know what rubber stamp means but I say do it (unless it just means approve which you already did?)!

PR #166, opened 2024-10-01
starting conditional analysis as a new file

annacuomo:
Super work in progress still

### DIFF
--- a/conditional_analysis.py
+++ b/conditional_analysis.py
@@ -1,0 +1,385 @@
+#!/usr/bin/env python3
+# pylint: disable=import-error,line-too-long
+
+"""
+Hail Batch workflow to perform conditional analysis using SAIGE-QTL
+"""
+
+import click
+import json
+import logging
+
+from datetime import datetime
+from os import getenv
+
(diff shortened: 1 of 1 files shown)


conditional_analysis.py
```diff
@@ -0,0 +1,423 @@
...
+    for celltype in celltypes:
+        logging.info(f'start summarising results for {celltype}')
+        summary_output_path = (
+            f'summary_stats/{celltype}_all_cis_{common_or_rare}_gene_level_results.tsv'
+        )
+
+        summarise_job = get_batch().new_python_job(
+            f'Summarise {common_or_rare} results for {celltype}'
+        )
+        summarise_job.depends_on(*celltype_jobs[celltype])
```
MattWellie, line 409:
```suggestion
        if celltype in celltype_jobs:
            summarise_job.depends_on(*celltype_jobs[celltype])
```

my guess is that by running it in fragments you didn't have any previous jobs in this batch. In a full end to end run you'd expect to find all the jobs, but not if the previous outputs exist when you set off this run.

PR #168, opened 2024-10-03
small tweaks

annacuomo:
Just a couple of tweaks from running it just now! https://batch.hail.populationgenomics.org.au/batches/492792

I didn't actually check with multiple output formats but I noticed the instructions were inconsistent between comma / space separated! And I find it more intuitive if the VDS object is actually called `.vds` but you may disagree!


annacuomo:
@Alexander-Stuckey used this PR to fix a couple of minor typos as well!

Alexander-Stuckey (approved):
Looks good to me, thanks for cleaning it up.

PR #170, opened 2024-10-04
Test config

annacuomo:
Properly access params from config file!


MattWellie (approved):
You're storing things in duplicate here... you're getting config[saige], then also getting config[saige][something else] which would be redundant. The code you have will work, but you might not need to get so much data for your config

annacuomo:
> You're storing things in duplicate here... you're getting config[saige], then also getting config[saige][something else] which would be redundant. The code you have will work, but you might not need to get so much data for your config

Yes sorry I did not appreciate the e.g. `saige.sv_test` was a subset of `saige`! If everything is already included in `saige` I'll remove the other lines? I'll make that change

MattWellie (approved):

PR #172, opened 2024-10-08
single combined sampleqc path

annacuomo:
Use combined tenk10k sample QC path instead of a TOB + BioHEART

- leaving the option of cohort-specific sample QC HT files commented out for now.
- updating files to use in the instructions at the top to most recent ones
- using sex_karyotype but converted to numeric coding

### DIFF
--- a/get_sample_covariates.py
+++ b/get_sample_covariates.py
@@ -106,12 +106,14 @@ def main(
     sample_qc_df = sample_qc_ht.to_pandas()
     # extract info and reformat to table
     sample_qc_df['sample_id'] = [str(s) for s in sample_qc_df['s']]
+    # go from sex karyotype (XX, XY) to sex numeric coding (2,1)
+    sample_qc_df['sex'] = sample_qc_df['sex_karyotype'].map({"XY": 1, "XX": 2})
     # only retain relevant columns
-    sex_df = sample_qc_df[['sample_id', 'sex_karyotype']]
+    sex_df = sample_qc_df[['sample_id', 'sex']]
 
     # add sex as unknown (0) if missing
     if fill_in_sex:
-        sex_df['sex_karyotype'] = sex_df['sex_karyotype'].fillna(0)
+        sex_df['sex'] = sex_df['sex'].fillna(0)
 
     # age
     # create a list from dictionary to populate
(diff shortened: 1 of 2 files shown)


get_sample_covariates.py
```diff
@@ -85,35 +79,35 @@ def main(
...
-    # sample_qc_df['sample_id'] = [str(s) for s in sample_qc_df['s']]
-    # # only retain relevant columns
-    # sex_df = sample_qc_df[['sample_id', 'sex']]
+    sample_qc_ht = hl.read_table(tenk10k_sampleqc_file_path)
+    # convert to pandas
+    sample_qc_df = sample_qc_ht.to_pandas()
+    # extract info and reformat to table
+    sample_qc_df['sample_id'] = [str(s) for s in sample_qc_df['s']]
+    # only retain relevant columns
+    sex_df = sample_qc_df[['sample_id', 'sex']]
```
katiedelange: This column field will need to be 'sex_karyotype' to use the genetically the inferred sex.

katiedelange (changes requested):
The path in the run command at the top is correct, but you will need to adjust the column you point to so that you use inferred sex rather than the (incomplete) reported sex column.

get_sample_covariates.py
annacuomo: Yup I just noticed and made the change!

get_sample_covariates.py
```diff
@@ -85,35 +79,39 @@ def main(
...
-    # # only retain relevant columns
-    # sex_df = sample_qc_df[['sample_id', 'sex']]
+    sample_qc_ht = hl.read_table(tenk10k_sampleqc_file_path)
+    # convert to pandas
+    sample_qc_df = sample_qc_ht.to_pandas()
+    # extract info and reformat to table
+    sample_qc_df['sample_id'] = [str(s) for s in sample_qc_df['s']]
+    # go from sex karyotype (XX, XY) to sex numeric coding (2,1)
+    sample_qc_df['sex'] = (
+        sample_qc_df['sex_karyotype'].map({"XY": 1, "XX": 2, "X": 0}).astype(int)
```
katiedelange, line 111: Looks good for now, but let's come back and generalise this to better capture more complex sex aneuploids.

annacuomo: I'll make an issue!

annacuomo: https://github.com/populationgenomics/saige-tenk10k/issues/173

PR #174, opened 2024-10-08
Add variant filtering options to the script

Alexander-Stuckey:
Add the ability to filter by a table of variants to the subsetting script.

Also fix the error strings for when a file exists.

### DIFF
--- a/subset_vds.py
+++ b/subset_vds.py
@@ -73,12 +73,12 @@ def check_output_already_exists(output_format: list[str], infile_name: str) -> N
             if to_path(
                 output_path(f"{infile_name}_subset", category="analysis")
             ).exists():
-                output_errors += f"The output VDS {infile_name}_subset.vds already exists. Refusing to overwrite it.\n"
+                output_errors += f"The output VDS {to_path(output_path(infile_name, category='analysis'))}_subset.vds already exists. Refusing to overwrite it.\n"
                 files_exist_errors = True
             if to_path(
                 output_path("subset_samples_file.txt", category="analysis")
             ).exists():
-                output_errors += f"The file {to_path(output_path('subset_samples_file.txt'))} exists. Refusing to overwrite it."
+                output_errors += f"The file {to_path(output_path('subset_samples_file.txt', category='analysis'))} exists. Refusing to overwrite it."
                 files_exist_errors = True
         if (
             format == "bed"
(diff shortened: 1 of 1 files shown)


subset_vds.py
```diff
@@ -66,28 +73,28 @@ def check_output_already_exists(output_format: list[str], infile_name: str) -> N
             if to_path(
                 output_path(f"{infile_name}_subset", category="analysis")
             ).exists():
-                output_errors += f"The output VDS {infile_name}_subset.vds already exists. Refusing to overwrite it.\n"
+                output_errors += f"The output VDS {to_path(output_path(infile_name, category='analysis'))}_subset.vds already exists. Refusing to overwrite it.\n"
```
MattWellie, line 76: it's not an expensive operation, but you're creating this Path object once to check it, and a second time to print it. You could create it as a variable instead of creating it twice.

subset_vds.py
```diff
@@ -197,8 +204,7 @@ def subset_by_samples(
     Returns:
         VariantDataset: the subset vds
     """
-    subset_vds: VariantDataset
-    subset_vds = filter_samples(input_vds, subset_sample_list)
+    subset_vds: VariantDataset = filter_samples(input_vds, subset_sample_list)
```
MattWellie, line 207: do you wanna check here that you keep at least one column? I've not bothered reading the script again in full, but I can't remember if you check that the sample subset is a true subset of the available samples in the VDS

MattWellie (approved):
potentially a change to be made confirming that your requested sample subset it present in the full sample set

subset_vds.py
Alexander-Stuckey: I had a check that one doesn't enter 0, or that `n_samples` is `<=` to the total, but I'll update the check to account for negative values of `n_sampels` as well.

subset_vds.py
Alexander-Stuckey: Hopefully this addresses it.

Alexander-Stuckey:
> potentially a change to be made confirming that your requested sample subset it present in the full sample set

The way I subset samples is that rather than specifying a list of samples, you instead just give a number and you get that many people from the original VDS

subset_vds.py
```diff
@@ -63,31 +70,33 @@ def check_output_already_exists(output_format: list[str], infile_name: str) -> N
     files_exist_errors: bool = False
     for format in output_format:
         if format == "vds":
-            if to_path(
+            if vds_path := to_path(
```
MattWellie, line 73: Gotta check your Walrus, son

```
if path := to_path('gs://cpg-acute-care-main/reanalysis/2024-10-03/config.toml').exists():
    print(path)
    
> True
if (path := to_path('gs://cpg-acute-care-main/reanalysis/2024-10-03/config.toml')).exists():
    print(path)
    
> gs://cpg-acute-care-main/reanalysis/2024-10-03/config.toml
```

If you're using a method on the thing your walrus operator creates, you need to bracket the creation of the object specifically

```if (variable := the thing you create).method():```

Alexander-Stuckey: Never used walri before, so good to learn that.

MattWellie (approved):
The collective noun for walruses might be "an ugly of walruses"
https://www.animalsandenglish.com/collective-nouns15.html

PR #175, opened 2024-10-08
drop NAs prior to permuting donors

annacuomo:
The next line permutes donor ids for calibration purposes, but the donor ids newly assigned may have missing values, and may be not present in the genotype files for association, so worth dropping rows with NA values at this stage.

Note: we have a strategy in place to fill in NAs for sex and age if needed, but not for PCs?


katiedelange:
I don't think you want to fill in NAs for PCs for any samples you are going to run association on? If there's missing PCs for samples you are including in the analysis (have WGS and scRNA-seq data, pass QC) then the PCA analysis should be rerun to include them.

katiedelange (approved):

PR #176, opened 2024-10-09
Extend the subset_vds.py to be able to subset on a sample list

katiedelange:
The main change here is adding an optional argument to filter the VDS using a space-separated list of sample IDs.

A few additional changes:
- Allow the user to adjust the suffix used for the output files (default:  "_subset")
- Genotype information should be stored in the base buckets, rather than analysis
- Fixed an arguments bug with filter_variants

### DIFF
--- a/subset_vds.py
+++ b/subset_vds.py
@@ -56,7 +56,9 @@ from hail.vds import (
 from hail.vds.variant_dataset import VariantDataset
 
 
-def check_output_already_exists(output_format: list[str], infile_name: str) -> None:
+def check_output_already_exists(
+    output_format: list[str], infile_name: str, suffix: str
+) -> None:
     """Check for existence of output files
 
     Args:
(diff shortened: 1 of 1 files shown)


subset_vds.py
```diff
@@ -335,14 +340,14 @@ def main(
         else:
             subset_vds = subset_by_variants(input_variants, keep_variants, input_vds)
 
-    if n_samples:
-        subset_sample_list = get_subset_sample_list(input_vds, n_samples)
+    if n_samples or sample_list:
+        subset_sample_list = sample_list or get_subset_sample_list(input_vds, n_samples)
```
Alexander-Stuckey, line 344: What happens here if someone provides both a sample list and `n_samples`? I suggest that we add a check earlier to only allow one.

Alexander-Stuckey (changes requested):
Suggest a check to allow only one of sample list or `n_samples`. The rest looks good to me (and thanks for fixing the issue with `filter_variants`)

subset_vds.py
```diff
@@ -385,16 +403,25 @@ def main(
...
+        and not args.sample_list
+        and not args.intervals
+        and not args.variant_table
+    ):
         raise AttributeError(
-            "If neither a number of samples or intervals are provided, this script will not do any subsetting!"
+            "If no filtering fields (n_samples, sample_list, intervals and/or variant_table) are provided, this script will not do any subsetting!"
         )
+    if args.n_samples and args.sample_list:
+        raise AttributeError("You can only provide one of n_samples or sample_list.")
```
katiedelange, line 416: I think this line was already catching that scenario?

Alexander-Stuckey (approved):
Looks good to me

PR #178, opened 2024-10-10
Use splitlines() instead of readlines() to prevent trailing newlines

katiedelange:
Using readlines() keeps a trailing '\n' at the end of each ID/intervl. Replacing this with .read().splitlines() fixes this, and should be tractable as we are unlikely to have extremely large files being inputted here.


annacuomo (approved):
looks good!

PR #181, opened 2024-10-11
Simplify get_gt_vcf checkpoints, add 1 more

MattWellie:
- drops all the complicated checkpointing logic. We can retrieve if required, but it could be obscuring something currently causing failures
- adds an additional post dense-mt checkpoint after we convert the VDS. This might not solve things, but should make iterations while troubleshooting faster


annacuomo (approved):
let's try it!

PR #183, opened 2024-10-11
Efficiency gains... maybe

MattWellie:
The Table returned from LD pruning isn't saved at any point - by saving it we should be able to annotate between the MT and HT more effectively

Reduces the number of count operations done on lazy-loading data from 3 to 1


annacuomo (approved):
LGTM

PR #184, opened 2024-10-12
Vds not vcf

annacuomo:
Ok so this was confusing, some of my make group file jobs were failing ([e.g.](https://batch.hail.populationgenomics.org.au/batches/494002/jobs/2385)) because they were not finding the locus in a chromosome, even though these were extracted by the same files.

I realised that the issue was what has been an annoying confusion here: we are using `GRCh38` as our genome reference, and Hail by default uses the `chr` nomenclature for chromosomes (so, e.g. `chr1`).
SAIGE-QTL however uses numerical only chromosomes (so, e.g. `1`) which is why we remove 'chr' from all genotype files ([bims](https://github.com/populationgenomics/saige-tenk10k/blob/main/get_genotype_vcf.py#L87) and [vcfs](https://github.com/populationgenomics/saige-tenk10k/blob/main/get_genotype_vcf.py#L76)).
Confusingly, this is also what Hail uses by default for `GRCh37` so here when opening the VCF file (with the removed `chr`) I got confused and opened it thinking it was reference 37.

In this PR I switch from using the chromosome specific VCF files to just re-opening the VDS so I make sure the reference is 38, add `chr` to the intervals before parsing and then I manually remove it gain just before saving to file, to please SAIGE-QTL.

I also noticed a typo in the formatting of the `bioheart` cohort as a covariate in the toml file so changing that also.

### DIFF
--- a/make_group_file.py
+++ b/make_group_file.py
@@ -84,7 +84,7 @@ def make_group_file(
         f'{allele[0]}:{allele[1]}' for allele in ds_result.alleles.collect()
     ]
     variants = [
-        f'{variants_chrom_pos[i]}:{variants_alleles[i]}'.replace('chr','')
+        f'{variants_chrom_pos[i]}:{variants_alleles[i]}'.replace('chr', '')
         for i in range(len(variants_chrom_pos))
     ]
 
--- a/make_group_file.py
+++ b/make_group_file.py
@@ -61,7 +61,7 @@ def make_group_file(
     num_chrom = gene_df.columns.values[0]
     window_start = gene_df.columns.values[1]
     window_end = gene_df.columns.values[2]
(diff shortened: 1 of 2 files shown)


MattWellie:
Lots of duplicate lines here, is this still a work in progress or can it be tidied up?

annacuomo:
> Lots of duplicate lines here, is this still a work in progress or can it be tidied up?

sorry,  just tidied up

annacuomo:
tested here: https://batch.hail.populationgenomics.org.au/batches/496946
files in `cpg-bioheart-test/saige-qtl/bioheart_n990_and_tob_n1055/241004_n100/input_files/group_files_test2/`

MattWellie (approved):

PR #192, opened 2024-10-15
Script for generating VDS -> Mt subsets

MattWellie:
Candidate script for pre-filtering to single-chrom MTs


annacuomo (changes requested):
Happy to do something like this, noting that pre-[this PR](https://github.com/populationgenomics/saige-tenk10k/pull/184) I was similarly using chrom-specific files.

If we want to do this rather than cleaning up the `chr` issue and reverting that change I'd probably want to do this properly and also restrict to rare variants?

annacuomo:
I think we can close this @MattWellie ?

PR #193, opened 2024-10-16
save chrom rare mts

annacuomo:
Rather than having a separate script as in https://github.com/populationgenomics/saige-tenk10k/pull/192 just save the MT here, which also 
* subsets to rare variants only, which is what I want (group files are only needed for the rare variant test)
* removes related individuals prior to running variant QC

Slack thread for context: https://centrepopgen.slack.com/archives/C030X7WGFCL/p1728950510597129

Easy solution to essentially reverting to the change prior to this PR https://github.com/populationgenomics/saige-tenk10k/pull/184 but avoiding the issue with the `chr` notation because the renaming only happens on the VCF file, and this saves the chrom + rare variant MT in parallel to exporting to VCF, prior to the renaming / indexing

### DIFF
--- a/get_genotype_vcf.py
+++ b/get_genotype_vcf.py
@@ -200,6 +200,22 @@ def main(
 
             mt = hl.variant_qc(mt)
 
+            if not rv_vcf_existence_outcome or rv_mt_existence_outcome:
+                # rare variants only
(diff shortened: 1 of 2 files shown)


MattWellie (changes requested):
Generally good, but adding #195 would make this more efficient

MattWellie (approved):
🤞

PR #194, opened 2024-10-16
Group file script without collecting over the distributed object

MattWellie:
Might be worth giving this a run in test and seeing if the syntax works out

The group_file script collects all loci and variants in the MT, then operates on them. This is a hugely inefficient operation, as it localises the data.

This is an attempt to do the same thing inside the mt:

1. annotate a new value into the MT 'chr:pos:ref:alt' (with "chr" prefix removed)
2. calculate distances and re-annotate into the MT
3. calculate the weights and re-annotate into the MT
4. export a table to TSV containing either `'variant_string', 'gene', 'distance', 'weight:dTSS'` or `'variant_string', 'gene'`

This would need some editing/transposition to exactly match the format you were generating previously, but it should scale much better than the fully-collected version

### DIFF
--- a/make_group_file.py
+++ b/make_group_file.py
@@ -136,8 +136,8 @@ def make_group_file(
         chrom_mt = chrom_mt.key_rows_by(chrom_mt.var).select_rows('gene').rows()
         categories_data = {'gene': [gene, gene, gene], 'category': ['var', 'anno']}
 
-    chrom_mt.export(group_file.replace('.tsv', '_tmp.tsv'))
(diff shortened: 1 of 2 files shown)


annacuomo (changes requested):
Ok I have tried to adjust the syntax based on testing it in a notebook, but definitely would be good to have a second pair of eyes on this

annacuomo:
testing to make sure output file looks ok: https://batch.hail.populationgenomics.org.au/batches/499346/jobs/1

annacuomo (changes requested):
the syntax of writing to a _tmp file and reading back from it does not work, can you help?

annacuomo (approved):

PR #203, opened 2024-10-21
Manual conditional analysis

annacuomo:
While we wait to see if a current issue with SAIGE-QTL's conditional analysis can be resolved ([PR](https://github.com/populationgenomics/saige-tenk10k/pull/166), latest [log](https://batch.hail.populationgenomics.org.au/batches/493996/jobs/3)), here is a workaround to still be able to run the pipeline after conditioning on a given genetic variant.

This adds two files
* one creating a conditional file -- mapping top variant to gene, so that we know what variant to condition on for each gene
* one creating a new pheno cov file (file containing expression of gene of interest plus all relevant covariates) that takes an existing one and adds the genotype of the variant we want to condition on and adds that as a column, so that it can be used as an additional covariate

It also modifies the `get_genotype_vcf.py` script to save common variants for each chromosome as an mt, similar to what we are already doing for rare variants to facilitate our make group file script (PR for context: https://github.com/populationgenomics/saige-tenk10k/pull/193)

-----

**Testing**

* Running `get_genotype_vcf.py` in test to get chr21 **`common`** mt: https://batch.hail.populationgenomics.org.au/batches/514577/jobs/1 🎉 
* the making of the conditional file had already been tested as part of the original attempt at conditional analysis (PR: https://github.com/populationgenomics/saige-tenk10k/pull/166), files here: `cpg-bioheart-test-analysis/saige-qtl/bioheart_n990_and_tob_n1055/241004_n100/output_files/conditioning_on_top_one_variant_from_cv_test_round1`
* running the add condition to pheno script: https://batch.hail.populationgenomics.org.au/batches/514594/jobs/1 🎉 

It worked, files here: `cpg-bioheart-test/saige-qtl/bioheart_n990_and_tob_n1055/241004_n100/input_files/pheno_cov_files_condition/B_naive/chr21`

PR #205, opened 2024-10-22
add option to only test some genes

annacuomo:
Because the new `manual` conditional analysis (https://github.com/populationgenomics/saige-tenk10k/pull/203) adds the SNP as a column in the covariates and the [covariates to use are specified in the TOML config](https://github.com/populationgenomics/saige-tenk10k/blob/main/saige_assoc_test.toml#L12-L13), it doesn't make sense in this case to run for a whole chromosome, so adding an option to just test specified genes here


Alexander-Stuckey (approved):
looks good to me

PR #206, opened 2024-10-23
saige is not happy with : in name

annacuomo:
I assume the issue [here](https://batch.hail.populationgenomics.org.au/batches/515517/jobs/2) is the `:` in the covariate, hopefully an easy fix 🤞


Alexander-Stuckey (approved):
Looks good to me

PR #207, opened 2024-10-23
add chr

annacuomo:
[new issue](https://batch.hail.populationgenomics.org.au/batches/516301/jobs/2) 🫠
I assume it's reading the '5' as a number, so adding `chr` before


MattWellie (approved):
sure, why not 🙃

PR #210, opened 2024-10-29
add option to test only specific genes

annacuomo:
same as https://github.com/populationgenomics/saige-tenk10k/pull/205 but for the common variant association pipeline


silkm (approved):
Looks good!

PR #214, opened 2024-11-26
Coloc

annacuomo:


coloc/coloc_runner.py
```diff
@@ -0,0 +1,246 @@
...
+        '''
+    eqtl_r = eqtl_r %>% filter(!is.na(beta))
+    eqtl_r = eqtl_r %>% distinct(snp, .keep_all = TRUE)
+    eqtl_r$varbeta = eqtl_r$se**2
+    eqtl_r$position = eqtl_r$pos
+    eqtl_r = eqtl_r %>% select(beta, varbeta, position, snp)
+
+    eqtl_r = eqtl_r %>% as.list()
+    eqtl_r$type = 'quant'
+    eqtl_r$sdY = 1
```
hopedisastro, line 80: ah, this I forgot - sdY is not necessarily = 1 for Poisson count data right?

coloc/coloc_runner.py
```diff
@@ -0,0 +1,246 @@
...
+        """
+        To avoid having too many jobs running at once, we have to limit concurrency.
+        """
+        if len(_dependent_jobs) >= max_parallel_jobs:
+            job.depends_on(_dependent_jobs[-max_parallel_jobs])
+        _dependent_jobs.append(job)
+
+    # read in gene annotation file
+    var_table = pd.read_csv(
+        'gs://cpg-bioheart-test/str/240_libraries_tenk10kp1_v2/concatenated_gene_info_donor_info_var.csv',
```
hopedisastro, line 159: update this with 300 library var file path

coloc/coloc_runner.py
```diff
@@ -0,0 +1,246 @@
...
+
+        # read in eGenes file
+        egenes_file = (
+            egenes_file
+        ) = f'{egenes_file_path}/{celltype}_all_cis_cv_gene_level_results.tsv'
+        result_df_cfm = pd.read_csv(
+            egenes_file,
+            sep='\t',
+        )
+        result_df_cfm = result_df_cfm[
```
hopedisastro, line 178: so right now you're doing coloc for every egene with a eQTL that passes the FDR threshold. 

This is slightly different from me: I only colocalise for every egene with an eQTL that passes FDR AND the eTR is fine-mapped and deemed candidate causal for driving the gene.

coloc/coloc_runner.py
annacuomo: right! Sorry I didn't really look at the R / coloc part of the code, what do you suggest I do here?

coloc/coloc_runner.py
annacuomo: yes I did notice that, and was going to ask you how quick this is? I can be more stringent with the significance threshold, e.g. 5e-6 or even 5e-8 on the smallest p-value for that gene, but have not (yet?) run fine-mapping atm

hopedisastro: it is quick to run; takes half a day to run on everything all cell types genome-wide

coloc/coloc_runner.py
```diff
@@ -0,0 +1,246 @@
...
+@click.option(
+    '--snp-gwas-file',
+    help='Path to the SNP GWAS file',
+    default='gs://cpg-bioheart-test/str/gwas_catalog/gcst/gcst-gwas-catalogs/GCST011071_parsed.tsv',
+)
+@click.option(
+    '--celltypes', help='Cell types to run, single str, comma separated', default='ASDC'
+)
+@click.option('--cis-window-size', help='Cis window size used', default=100000)
+@click.option('fdr-threshold', help='FDR threshold', default=0.05)
```
hopedisastro, line 126: this is throwing the error

hopedisastro, line 126:
```suggestion
@click.option('--fdr-threshold', help='FDR threshold', default=0.05)
```

coloc/coloc_runner.py
annacuomo: I've added N + MAF for now as suggested https://chr1swallace.github.io/coloc/articles/a02_data.html#what-if-i-dont-have-sdy

coloc/coloc_runner.py
annacuomo: ah oops sorry didn't see this! Changed that below, would have saved me a couple of minutes 😅

coloc/coloc_runner.py
```diff
@@ -0,0 +1,251 @@
...
+        eqtl_r = ro.conversion.get_conversion().py2rpy(eqtl)
+    ro.globalenv['eqtl_r'] = eqtl_r
+    ro.globalenv['gene'] = gene
+    ro.r(
+        '''
+    eqtl_r = eqtl_r %>% filter(!is.na(beta))
+    eqtl_r = eqtl_r %>% distinct(snp, .keep_all = TRUE)
+    eqtl_r$varbeta = eqtl_r$se**2
+    eqtl_r$position = eqtl_r$pos
+    eqtl_r = eqtl_r %>% select(beta, varbeta, position, snp, N, MAF)
```
katiedelange, line 80: Just confirming, N is a field that's provided for each site in your eQTL input file?

coloc/coloc_runner.py
```diff
@@ -0,0 +1,251 @@
...
+        """
+        To avoid having too many jobs running at once, we have to limit concurrency.
+        """
+        if len(_dependent_jobs) >= max_parallel_jobs:
+            job.depends_on(_dependent_jobs[-max_parallel_jobs])
+        _dependent_jobs.append(job)
+
+    # read in gene annotation file
+    var_table = pd.read_csv(
+        'gs://cpg-bioheart-test/saige-qtl/300-libraries/combined_anndata_obs_vars/300_libraries_concatenated_harmony_filtered_vars.csv',
```
katiedelange, line 161: This file is quite confusing, but if I understand correctly you're only using it to set the coordinates for the windows and it's ok that it's the same for every celltype etc? Ideally this should probably get swapped out with a parameter at some point.

katiedelange (approved):
LGTM

coloc/coloc_runner.py
annacuomo: Yep, N is one out of outputs of SAIGE-QTL! For the test example file I am using:
Screenshot 2024-11-28 at 3 31 50 PM

annacuomo: but this is a good reminder to check with Wei how exactly this and AC / AF are calculated

coloc/coloc_runner.py
annacuomo: This is just the `adata.var` object: the sc expression file comes in [AnnData format](https://anndata.readthedocs.io/en/stable/) with the count matrix as the main info, and `obs` and `var` tables including information on cells and genes (only) respectively.

Here I have just extracted the `var` object the info in which mainly comes from Gencode.

But yes to making this a parameter, I'll do this now

annacuomo: https://github.com/populationgenomics/saige-tenk10k/pull/214/commits/b0b93332e01e7749c47e9c63abcab863d9942bb7

PR #221, opened 2024-12-10
pcs file path as argument

annacuomo:
Hey @katiedelange, can you please check the paths in the instructions at the top? And if you spot any other major issues


katiedelange (approved):
Slight edit to the PCs to use the ones calculated on the QC-pass European subset only (the other version includes additional samples of diverse ancestry).

PR #222, opened 2024-12-10
Split get genotypes

annacuomo:
As noted in [this thread](https://centrepopgen.slack.com/archives/C030X7WGFCL/p1733815889312609?thread_ts=1732046164.229849&cid=C030X7WGFCL) the second part of this script is potentially more time consuming and expensive, so splitting into two parts here to split the debug

### DIFF
--- a/get_genotype_vcf.py
+++ b/get_genotype_vcf.py
@@ -29,14 +29,11 @@ import pandas as pd
 
 import hail as hl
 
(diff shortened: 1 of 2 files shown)


MattWellie (approved):
looks fine to me, it's a copy-paste to split the two separate sets of functionality out of the one file so should be pretty straightforward

PR #223, opened 2024-12-10
Force checkpoint overwrite

MattWellie:
Allow continuation from a partially written checkpoint


annacuomo (approved):
looks good to me, thanks!

PR #232, opened 2024-12-14
Reduce overall number of dependencies

MattWellie:
The previous change split off the summary step from the processing steps.

This means that the celltype_jobs dictionary was never used, so we can safely delete that. 

---

This change also introduces a slightly different behaviour - instead of setting a concurrency limit across all chromosomes/celltypes, jobs are now only dependent on other jobs on the same chromosome. This should hopefully remove some really extended dependency chains, which _could_ help with the concurrency limit/recursion issues

### DIFF
--- a/saige_assoc.py
+++ b/saige_assoc.py
@@ -274,6 +274,15 @@ def create_second_job(vcf_path: str) -> hb.batch.job.Job:
     return second_job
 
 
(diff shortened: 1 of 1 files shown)


annacuomo (approved):
🤞

annacuomo:
https://batch.hail.populationgenomics.org.au/batches/533466/jobs/1

PR #234, opened 2024-12-23
use round

annacuomo:
Because the cis window file read here is a single line, it gets read as the header of the data frame when opened in.
In the unlikely case where two values are identical (it happens [here](https://batch.hail.populationgenomics.org.au/batches/577759/jobs/95) when opening `cpg-tenk10k-main/saige-qtl/tenk10k-genome-2-3-eur/input_files/241210/cis_window_files/chr1/ENSG00000238009_100000bp.tsv` where both `chr` and `window_start` are 1), to avoid duplicated column names the second `1` gets turned into `1.1` which is then confusing when restricting an MT to a genomic range, as 1.1 is not a valid position. A bit hacky but "rounding" the value to the closest int fixes it here (in all cases `.1` would get rounded down).

screenshot of my notebook to illustrate the issue

Screenshot 2024-12-23 at 4 47 31 PM

### DIFF
--- a/make_group_file.py
+++ b/make_group_file.py
@@ -77,7 +77,7 @@ def make_group_file(
     print(f'gene file: {gene_file}')
     gene_df = pd.read_csv(gene_file, sep='\t')
     num_chrom = gene_df.columns.values[0]
-    window_start = gene_df.columns.values[1]
+    window_start = round(float(gene_df.columns.values[1]))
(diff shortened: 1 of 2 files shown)


hopedisastro (approved):

annacuomo:
https://batch.hail.populationgenomics.org.au/batches/580127/jobs/1

PR #235, opened 2024-12-31
Update and add multi pheno runner

annacuomo:
This PR adds a multi pheno + multi celltype runner, and removes all constants making them adjustable parameters instead:

* MAF threshold (esp relevant as we accidentally ran for variants with MAF<1%)
* GWAS threshold to determine whether to run coloc

It also updates default paths to latest freeze.

Multi pheno runner syntax worked in `test` and ran successfully: 

* [585031](https://batch.hail.populationgenomics.org.au/batches/585031/jobs/1) 
* [585032](https://batch.hail.populationgenomics.org.au/batches/585032/jobs/1) 

files in `cpg-tenk10k-test-analysis/saige-qtl/bioheart_n990_and_tob_n1055/241004_n100/coloc-snp-only/sig_genes_only`

### DIFF
--- a/coloc/coloc_runner.py
+++ b/coloc/coloc_runner.py
@@ -34,7 +34,9 @@ from cpg_utils import to_path
 from cpg_utils.hail_batch import get_batch, image_path, output_path
 
 
-def coloc_runner(gwas, eqtl_file_path, celltype, coloc_results_file):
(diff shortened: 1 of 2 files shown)


MattWellie (approved):
LGTMatt

PR #239, opened 2025-01-06
retain qc passing variants only

annacuomo:
The original plan here was that the VDS would be already filtered to QC passing variants, but that is not the case, so adding similar lines to the [genotype processing script ones](https://github.com/populationgenomics/saige-tenk10k/blob/main/get_genotype_vcf.py#L173-L174) to only keep QC pass variants when counting up.

Reminder that this script's only goal is to count the total number of variants we have split by frequency, for the purpose of including these numbers in our manuscript.


MattWellie (approved):

annacuomo:
```
analysis-runner \
    --description "count common, low freq and rare variant VCFs" \
    --dataset "tenk10k" \
    --access-level "full" \
    --output-dir "saige-qtl/" \
    python3 variant_counter.py --vds-path=gs://cpg-tenk10k-main/vds/tenk10k-genome-2-0.vds \
        --donors-to-keep=gs://cpg-tenk10k-main/large_cohort/tenk10k-genome-2-3-eur/sample_qc.ht \
        --donors-to-exclude=gs://cpg-tenk10k-main/large_cohort/tenk10k-genome-2-3-eur/all_wgs_and_scrnaseq_qc_fail_samples.ht \
        --variants-to-keep=gs://cpg-tenk10k-main/large_cohort/tenk10k-genome-2-3-eur/variants_qc.ht
```
https://batch.hail.populationgenomics.org.au/batches/588559/jobs/1

PR #240, opened 2025-01-16
only keep genes to test in chrom

annacuomo:
hoping this will avoid looking for input files in the wrong chrom


MattWellie (approved):
makes sense to me

annacuomo:
testing: https://batch.hail.populationgenomics.org.au/batches/589043/jobs/1

PR #250, opened 2025-10-02
update main invocation

annacuomo:


MattWellie (approved):
Looks good, hoping the test run works!

annacuomo:
the latest error seems connected with a different between quoted strings and unquoted.
https://batch.hail.populationgenomics.org.au/batches/633550/jobs/2

When I pass arguments I mostly include them via toml, eg in this case I had in the toml the line:
annotation_in_groupTest = 'functional'

but I think this leads SAIGE to look for the string 'functional' rather than just functional as it is in the group file.
I don't know if this is the solution but trying to pass these using click instead

PR #251, opened 2025-10-03
explode annos

annacuomo:
I had interpreted this wrong, I am not allowed to have more than one annotation per variant in the same row, they need to be exploded out instead.
Made one example in notebook and it was successful: https://batch.hail.populationgenomics.org.au/batches/633595
output file here: cpg-tenk10k-main-analysis/saige-qtl/tenk10k-genome-2-3-eur/output_files/251002/annotated/summary_stats/CD4_TCM_all_cis_rv_set_test_results.tsv

documentation: https://saigegit.github.io/SAIGE-doc/docs/set_step2.html

### DIFF
--- a/make_group_file.py
+++ b/make_group_file.py
@@ -18,7 +18,7 @@ analysis-runner \
     --description "make variant group input files" \
     --dataset "tenk10k" \
     --access-level "full" \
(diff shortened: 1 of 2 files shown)


MattWellie (approved):
sure!

PR #254, opened 2025-10-05
add specific anno to summary file also

annacuomo:
This may fix the summarising step errors? https://batch.hail.populationgenomics.org.au/batches/655813

### DIFF
--- a/saige_assoc_set_test.py
+++ b/saige_assoc_set_test.py
@@ -495,7 +495,7 @@ def main(
     # summarise results (per cell type)
     for celltype in celltypes:
         logging.info(f'start summarising results for {celltype}')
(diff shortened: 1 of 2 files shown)


MattWellie (approved):
yup, I think that's it 🤦 the `gene_results_path` bit in particular, as that sets the search path

annacuomo:
> Yeah that sounds right. We can either force the stage to run in this one situation, or we can delete one or more of the result files so this will naturally rerun. My preference would be for the latter, because forcing it to rerun would only be relevant in this one situation

Sounds good @MattWellie, PR here pretty please: https://github.com/populationgenomics/transfer-private/pull/205

PR #255, opened 2025-10-06
Groups open chromatin

annacuomo:
_very_ annoyingly the current set up of making group files (one per gene) with _all_ of the annotations and then only test for one at the time does not work.

I got lucky the first time (running with `--group-annos='functional'functional`) because *all* genes had at least one variant annotated as functional, but when I tried with e.g. `open_CD4_TCM` I get a bunch of fails because the group file exists and everything so the script starts to run but when subsetting to the relevant there are no variants left for many genes and SAIGE can't run.

If you have any suggestions more than welcome @MattWellie but at the moment I am just creating a bunch more group files, one per anno and then I'll have to add something to the saige_assoc_set_test.py script to skip if the group file required does not exist

EDIT: actually ideally we could at least save multiple files within the make_group_file function if we had multiple `required_anno(s)`, but the file naming would have to move from outside to inside the function I think?

### DIFF
--- a/make_group_file.py
+++ b/make_group_file.py
@@ -288,6 +288,7 @@ def make_group_file(
     start_idx = rows_matching.index[0]
 
     block_df = final_df.loc[start_idx : start_idx + 2].copy()
+    block_df = block_df.dropna(axis=1)
 
     with group_file.open('w') as gdf:
         block_df.to_csv(gdf, index=False, header=False, sep=' ')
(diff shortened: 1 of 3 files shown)


annacuomo:
I have made these files for `ENSG00000054611`
e.g. `cpg-tenk10k-main/saige-qtl/tenk10k-genome-2-3-eur/input_files/241210/group_files_mt_n1925/functional_and_open/functional/chr22/ENSG00000054611_100000bp_dTSS_weights.tsv`
in a notebook, and am testing here: https://batch.hail.populationgenomics.org.au/batches/656265

PR #256, opened 2025-10-07
avoid writing to file if block is empty

annacuomo:
🤦‍♀️ we know some of these blocks won't exist so should've expected this

context: https://github.com/populationgenomics/saige-tenk10k/pull/255

I feel like `exit` is maybe not quite the right syntax? 😆 

Not too huge an issue as there are no dependencies so all the genes / jobs that _can_ run still will

### DIFF
--- a/make_group_file.py
+++ b/make_group_file.py
@@ -285,13 +285,15 @@ def make_group_file(
     rows_matching = final_df[final_df.iloc[:, 0] == block_id]
 
     # Get the row index of the first matching row
-    start_idx = rows_matching.index[0]
-
-    block_df = final_df.loc[start_idx : start_idx + 2].copy()
-    block_df = block_df.dropna(axis=1)
-
-    with group_file.open('w') as gdf:
-        block_df.to_csv(gdf, index=False, header=False, sep=' ')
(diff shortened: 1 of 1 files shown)


make_group_file.py
```diff
@@ -285,13 +285,15 @@ def make_group_file(
...
-        block_df.to_csv(gdf, index=False, header=False, sep=' ')
+    if not rows_matching.empty:
+        start_idx = rows_matching.index[0]
+        block_df = final_df.loc[start_idx : start_idx + 2].copy()
+        block_df = block_df.dropna(axis=1)
+        with group_file.open('w') as gdf:
+            block_df.to_csv(gdf, index=False, header=False, sep=' ')
+    else:
+        print(f"Block '{block_id}' not found.")
+        exit()
```
MattWellie, line 296:
```suggestion
```

Personally I'd dump the `exit` here. It's the last statement in the method, and the method is called in it's own job. If you do nothing, it'll just end without producing any output file.

`exit` is an alias for `quit`, which throws an error on the way out. That could cancel other jobs in the batch. Probably not given that there are no dependencies, but it would mean the batch has a mix of successful and failed jobs

PR #261, opened 2025-10-16
Simplify summary process by removing Pandas

MattWellie:
The current summary process:
- reads each TSV file (1-5) lines into a dataframe
- adds the dataframe to a list
- concatenates all DFs
- writes unified DF as a CSV

for [this batch](https://batch.hail.populationgenomics.org.au/batches/1022845) this is 10754 separate results files all being processed as a dataframe, then requiring a monster merge. This WF has taken a day and a half with no obvious signs of progress :/

This PR suggests a simplification:
- open a local csv file
- read each file in turn
- if it's the first file in the cycle, use all lines (including the header), converting tabs to spaces. Write to the local file
- if it's !first file, only take the data lines from the file, converting tabs to spaces, write to the local file
- once we've finished iterating, open a final output file in GCP, write all data as a single operation

This does all the same things as the previous process, reading each file in turn, converting TSV -> CSV, and writing a new file, but it doesn't need Pandas to do it, and doesn't require a 10000+ DataFrame concatenation. 

Seems to work from local testing 😬

### DIFF
--- a/saige_assoc_set_test.py
+++ b/saige_assoc_set_test.py
@@ -262,7 +262,9 @@ def summarise_rv_results(
             if non_empty_files % 100 == 0:
                 print(f'Progress: {non_empty_files} processed, {empty_files} skipped.')
 
-            contents = file.open().readlines()
+            with file.open() as f:
+                contents = f.readlines()
+
             if first_file:
                 first_file = False
                 use_index = 0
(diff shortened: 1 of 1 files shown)


annacuomo (approved):
I can't say I fully understand but the rationale seems sensible, let's try it :)

